PR #4, opened 2020-09-17
Clean up Homie crate ready to publish

qwandor-google:


alsuren (approved):
Looks good. I would say that homie-device is probably okay to publish (although I suspect that we will need to break things when we change the error type).

I don't think I'd be comfortable publishing any of the others on crates.io yet, but the changes look good.

PR #5, opened 2020-09-17
make run.sh use systemd by default

alsuren:


run.sh
```diff
@@ -9,21 +9,44 @@ TARGET=${TARGET:-armv7-unknown-linux-gnueabihf}
...
+Type=simple
+User=pi
+WorkingDirectory=/home/pi
+Environment=RUST_BACKTRACE=1
+ExecStart=/home/pi/publish-mqtt
+Restart=always
+
+[Install]
+WantedBy=multi-user.target
+"
```
qwandor-google, line 42: This seems like it should be in a separate file, which we could also use at some point in the future to build release packages.

qwandor-google (approved):

PR #6, opened 2020-09-19
Refactor Mijia crate a bit

qwandor-google:
Just moving things around and adding a few doc comments.

### DIFF
--- a/mijia/src/decode.rs
+++ b/mijia/src/decode.rs
@@ -1,0 +1,83 @@
+use std::cmp::max;
+use std::convert::TryInto;
+use std::fmt::{Display, Formatter};
+
+/// A set of readings from a Mijia sensor.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Readings {
+    /// Temperature in ºC, with 2 decimal places of precision
+    pub temperature: f32,
+    /// Percent humidity
+    pub humidity: u8,
+    /// Voltage in millivolts
+    pub battery_voltage: u16,
+    /// Inferred from `battery_voltage` with a bit of hand-waving.
+    pub battery_percent: u16,
+}
+
+impl Display for Readings {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        write!(
+            f,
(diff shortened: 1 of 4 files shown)


publish-mqtt/src/main.rs
```diff
@@ -483,3 +482,23 @@ async fn handle_bluetooth_event(
 
     Ok(())
 }
+
+/// Read the given file of key-value pairs into a hashmap.
+/// Returns an empty hashmap if the file doesn't exist, or an error if it is malformed.
+pub fn hashmap_from_file(filename: &str) -> Result<HashMap<String, String>, io::Error> {
```
alsuren, line 488: (observation) I probably would have put it a bit closer to `run_sensor_system()` (it's the first thing that run_sensor_system() does, so probably immediately below) but in the long run we will probably split this crate out into modules too, so it doesn't really matter where it goes for now.

mijia/src/decode.rs
```diff
@@ -0,0 +1,83 @@
...
+            self.temperature, self.humidity, self.battery_voltage, self.battery_percent
+        )
+    }
+}
+
+impl Readings {
+    /// Decode the readings from the raw bytes of the Bluetooth characteristic value, if they are
+    /// valid.
+    /// Returns `None` if the value is not valid.
+    pub fn decode(value: &[u8]) -> Option<Readings> {
```
alsuren, line 32: long-term, I would be tempted to implement https://doc.rust-lang.org/std/convert/trait.TryFrom.html , and then use `let readings = value.try_into().ok()?` in the one place where it's used.

For now, I don't think that consumers will ever have `&[u8]` in their hand, so maybe `pub(crate)` and leave everything else as-is?

alsuren (approved):

publish-mqtt/src/main.rs
qwandor-google: Done.

mijia/src/decode.rs
qwandor-google: Done the latter.

PR #8, opened 2020-09-21
Mijia refactoring

qwandor-google:
I've tried to separate the Mijia-specific code from the general BLE stuff. I've also added a few comments.


mijia/src/bluetooth.rs
```diff
@@ -0,0 +1,201 @@
...
+                    // Variant(...)
+                    .as_iter()?
+                    .next()?
+                    // InternalDict(...)
+                    .as_iter()?
+                    .tuples::<(_, _)>()
+                    .filter_map(|(k, v)| {
+                        let k = k.as_str()?.into();
+                        let v = v.box_clone();
+                        let v = cast::<Vec<u8>>(&v)?.clone();
```
alsuren, line 160: I don't trust that this cast call is going to work when it encounters real data. I would be more comfortable if we had a test. Maybe deploy it to cottagepi and see if it works before merging?

I suppose I should have made a test for this as well.

mijia/src/bluetooth.rs
```diff
@@ -0,0 +1,201 @@
...
+use futures::FutureExt;
+use itertools::Itertools;
+use std::collections::HashMap;
+use std::error::Error;
+use std::fmt::{self, Display, Formatter};
+use std::str::FromStr;
+use std::sync::Arc;
+
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct DeviceId {
```
alsuren, line 19: probably wants a doc comment with the word opaque in it somewhere.

(observation) I wonder if id is the right name. I'm not used to "id"s being opaque structs with private fields. You have more experience of languages with private fields than me, so I will defer to your judgement on this. I guess we can't use Handle because we're already using that for JoinHandles and it would become confusing.

mijia/src/lib.rs
```diff
@@ -17,85 +18,127 @@ const CONNECTION_INTERVAL_CHARACTERISTIC_PATH: &str = "/service0021/char0045";
...
-                // InternalDict(...)
-                .as_iter()?
-                .tuples::<(_, _)>()
-                .filter_map(|(k, v)| Some((k.as_str()?.into(), v)))
-                .collect();
+// TODO before publishing to crates.io: annotate this enum as non-exhaustive.
+/// An event from a Mijia sensor.
+#[derive(Clone)]
+pub enum MijiaEvent {
+    // FIXME: stop using object_path as primary key. Can we think of something better?
```
alsuren, line 31:
```suggestion
```

mijia/src/lib.rs
```diff
@@ -17,85 +18,127 @@ const CONNECTION_INTERVAL_CHARACTERISTIC_PATH: &str = "/service0021/char0045";
...
+    fn from(conn_msg: Message) -> Option<Self> {
+        match BluetoothEvent::from(conn_msg) {
+            Some(BluetoothEvent::Value { object_path, value }) => {
+                // TODO: Make this less hacky.
+                let object_path = object_path
+                    .strip_suffix(SENSOR_READING_CHARACTERISTIC_PATH)?
+                    .to_owned();
+                let readings = Readings::decode(&value)?;
+                Some(MijiaEvent::Readings {
+                    id: DeviceId { object_path },
```
alsuren, line 46: maybe make a `DeviceId::new(&str)` and do the clone in there?

mijia/src/lib.rs
```diff
@@ -17,85 +18,127 @@ const CONNECTION_INTERVAL_CHARACTERISTIC_PATH: &str = "/service0021/char0045";
...
+                    None
+                }
+            })
+            .collect();
+        Ok(sensors)
+    }
+
+    /// Assuming that the given device ID refers to a Mijia sensor device and that it has already
+    /// been connected, subscribe to notifications of temperature/humidity readings, and reduce the
+    /// connection interval to save power.
```
alsuren, line 99:
```suggestion
    /// been connected, subscribe to notifications of temperature/humidity readings, and tweak the
    /// connection interval to save power.
```
I think we're increasing the interval, aren't we?

mijia/src/lib.rs
```diff
@@ -17,85 +18,127 @@ const CONNECTION_INTERVAL_CHARACTERISTIC_PATH: &str = "/service0021/char0045";
...
+    /// Notifications will be delivered as events by `MijiaSession::event_stream()`.
+    pub async fn start_notify_sensor(&self, id: &DeviceId) -> Result<(), anyhow::Error> {
+        let temp_humidity_path = id.object_path.to_string() + SENSOR_READING_CHARACTERISTIC_PATH;
+        let temp_humidity = dbus::nonblock::Proxy::new(
+            "org.bluez",
+            temp_humidity_path,
+            DBUS_METHOD_CALL_TIMEOUT,
+            self.session.connection.clone(),
+        );
+        temp_humidity.start_notify().await?;
```
alsuren, line 110: your bluetooth.rs split is such a good idea that I'm suddenly trying to think of ways to move things like this into bluetooth.rs (definitely not in this PR).

publish-mqtt/src/main.rs
```diff
@@ -301,11 +298,12 @@ async fn bluetooth_connection_loop(
 async fn check_for_sensors(
     state: Arc<Mutex<SensorState>>,
     bt_session: &MijiaSession,
-    sensor_names: &HashMap<String, String>,
+    sensor_names: &HashMap<MacAddress, String>,
 ) -> Result<(), anyhow::Error> {
-    bt_session.start_discovery().await?;
+    bt_session.session.start_discovery().await?;
```
alsuren, line 303:
```suggestion
    session.bt_session.start_discovery().await?;
```

publish-mqtt/src/main.rs
```diff
@@ -332,7 +330,11 @@ async fn connect_first_sensor_in_queue(
     sensors_connected: &mut Vec<Sensor>,
     sensors_to_connect: &mut VecDeque<Sensor>,
 ) -> Result<(), anyhow::Error> {
-    println!("{} sensors in queue to connect.", sensors_to_connect.len());
+    println!(
+        "{} sensors in queue to connect, {} connected.",
+        sensors_to_connect.len(),
+        sensors_connected.len()
+    );
```
alsuren, line 337: ❤️

mijia/src/bluetooth.rs
alsuren: I put it on cottagepi with some debugging and debugs as `Variant(InternalArray { data: [48, 88, 91, 5, 1, 23, 33, 215, 56, 193, 164, 40, 1, 0], inner_sig: Signature("y") })` but then casts to None.

mijia/src/lib.rs
qwandor-google: You're quite right.

publish-mqtt/src/main.rs
qwandor-google: Done.

mijia/src/lib.rs
qwandor-google: Done.

mijia/src/bluetooth.rs
qwandor-google: Added a doc comment. Not sure what the best name is, I agree that it's a bit weird for something opaque to be called an ID.

alsuren: It turns out that TypeId [from the standard library](https://doc.rust-lang.org/src/core/any.rs.html#417-419) is defined as:
```
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
#[stable(feature = "rust1", since = "1.0.0")]
pub struct TypeId {
    t: u64,
}
```
so I think we're probably doing the rustic thing here.

mijia/src/bluetooth.rs
```diff
@@ -164,7 +166,7 @@ impl BluetoothSession {
                     .filter_map(|(k, v)| {
                         let k = k.as_str()?.into();
                         let v = v.box_clone();
-                        let v = cast::<Vec<u8>>(&v)?.clone();
+                        let v = cast::<Variant<Vec<u8>>>(&v)?.0.clone();
```
alsuren, line 169: Pre-approved for when you push the change to use into_iter() etc. I really think that the best time to fix it properly is as part of #9 so I'm happy to leave it as-is for now.

If you want to write a test that's realistic, you will probably need to add an extra layer of nesting, so you can't Any your way out of trouble without boxed_clone, (we're currently a hashmap of {sv}, where the v is a Variant of &[u8] or something, so we're already borrowing from the outer hashmap)

alsuren (approved):

PR #10, opened 2020-09-23
Refactor SensorState to have a single set of sensors, and use states

### LINKED ISSUE #1: Refactor `SensorState::sensors_*` into a single `SensorState::sensors`
... and a `::next_sensor_idx` pointer for scheduling connection attempts/staleness checks.

There is a sketch of this here: https://github.com/alsuren/reading-xiaomi-temp/pull/32

qwandor-google:
Moving here instead of alsuren/reading-xiaomi-temp#32.

Fixes #1.

### DIFF
--- a/publish-mqtt/src/main.rs
+++ b/publish-mqtt/src/main.rs
@@ -390,13 +390,13 @@ async fn connect_sensor_at_idx(
     session: &MijiaSession,
     idx: usize,
 ) -> Result<(), anyhow::Error> {
+    let mut sensor = clone_sensor_at_idx(state.clone(), idx).await;
     {
         state.lock().await.sensors[idx].connection_status = ConnectionStatus::Connecting {
             reserved_until: Instant::now() + SENSOR_CONNECT_RESERVATION_TIMEOUT,
         }
     }
     // Try to connect to a sensor.
-    let mut sensor = clone_sensor_at_idx(state.clone(), idx).await;
     println!("Trying to connect to {}", sensor.name);
     let status = connect_start_sensor(session, &mut sensor).await;
     let mut state = state.lock().await;
--- a/publish-mqtt/src/main.rs
+++ b/publish-mqtt/src/main.rs
@@ -392,7 +392,10 @@ async fn connect_sensor_at_idx(
     let mut state = state.lock().await;
     match status {
         Err(e) => {
-            println!("Failed to connect to {}: {:?}", sensor.name, e);
+            println!(
+                "Failed to connect to {} (now {:?}): {:?}",
+                sensor.name, sensor.connection_status, e
+            );
         }
         Ok(()) => {
             println!("Connected to {} and started notifications", sensor.name);
--- a/publish-mqtt/src/main.rs
+++ b/publish-mqtt/src/main.rs
@@ -266,6 +266,7 @@ async fn bluetooth_connection_loop(
 ) -> Result<(), anyhow::Error> {
     let mut next_scan_due = Instant::now();
     loop {
+        // Print count and list of sensors in each state.
         {
             let counts = state
                 .lock()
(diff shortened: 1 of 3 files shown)


publish-mqtt/src/main.rs
```diff
@@ -363,20 +424,13 @@ async fn connect_start_sensor<'a>(
         .await
         .with_context(|| std::line!().to_string())?;
```
qwandor-google, line 425: Should the `connection_status` be updated if this fails, rather than leaving it in the `Connecting` state?

publish-mqtt/src/main.rs
```diff
@@ -363,20 +424,13 @@ async fn connect_start_sensor<'a>(
...
-                .with_context(|| std::line!().to_string())?;
-            sensor.connection_status = ConnectionStatus::Connected;
-            sensor.last_update_timestamp = Instant::now();
-            Ok(())
-        }
+        Ok(()) => Ok(()),
         Err(e) => {
             // If starting notifications failed a second time, disconnect so
             // that we start again from a clean state next time.
             match sensor.connection_status {
```
qwandor-google, line 431: Is the `connection_status` ever anything other than `Connecting` here?

alsuren: Hrm. I guess not. This suggests that we want to do the retrying here until SENSOR_CONECT_LOCK_TIMEOUT/2 or something and kill off the failed once state? I think this is something I wanted to do anyway eventually.

publish-mqtt/src/main.rs
```diff
@@ -397,30 +451,28 @@ async fn connect_start_sensor<'a>(
...
 }
 
-/// If a sensor hasn't sent any updates in a while, disconnect it and add it back to the
-/// connect queue.
-async fn disconnect_first_stale_sensor(
-    homie: &mut HomieDevice,
-    sensors_connected: &mut Vec<Sensor>,
-    sensors_to_connect: &mut VecDeque<Sensor>,
+/// If the sensor hasn't sent any updates in a while, disconnect it so we will try to reconnect.
+async fn check_for_stale_sensor(
```
alsuren, line 455: If we fix the to-do then this could be called disconnect_if_stale()?

publish-mqtt/src/main.rs
```diff
@@ -309,51 +364,57 @@ async fn check_for_sensors(
...
+    idx: usize,
 ) -> Result<(), anyhow::Error> {
-    println!(
-        "{} sensors in queue to connect, {} connected.",
-        sensors_to_connect.len(),
-        sensors_connected.len()
-    );
+    {
+        state.lock().await.sensors[idx].connection_status = ConnectionStatus::Connecting {
+            reserved_until: Instant::now() + Duration::from_secs(5 * 60),
```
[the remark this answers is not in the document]
qwandor-google: Done.

publish-mqtt/src/main.rs
qwandor-google: Fixed by moving the `clone_sensor_at_idx` call in `connect_sensor_at_idx` higher up, so it clones the sensor before changing the state of the original to `Connecting`.

publish-mqtt/src/main.rs
qwandor-google: I think this is now unnecessary because the state will revert to what it was before it was updated to `Connecting`. Now I think the only reason the `Connecting` state has an effect is if an update for the sensor comes in before this function returns. We can probably solve this better with locking.

publish-mqtt/src/main.rs
```diff
@@ -309,51 +364,57 @@ async fn check_for_sensors(
...
-                println!("Failed to connect to {}: {:?}", sensor.name, e);
-                sensors_to_connect.push_back(sensor);
-            }
-            Ok(()) => {
-                println!("Connected to {} and started notifications", sensor.name);
-                sensors_connected.push(sensor);
-            }
+    let mut sensor = clone_sensor_at_idx(state.clone(), idx).await;
+    println!("Trying to connect to {}", sensor.name);
+    let status = connect_start_sensor(session, &mut sensor).await;
```
alsuren, line 403: I'm a bit sad that connect_start_sensor() mutates sensor rather than just returning the new connection_status, but think let's merge this and I'll have a crack at it over the weekend.

alsuren (approved):

PR #11, opened 2020-09-26
move subscription retrying into connect_and_subscribe_sensor(); simplify

alsuren:
This allows us to delete the SubscribingFailedOnce state, because it is folded into Connecting.

There are also some meanderings about using id rather than idx to identify sensors. Once we stop caring about the ConnectionStatus while connecting. Mijia only cares about the id, so if we can also only care about the id then we are only holding a single piece of state across the connection attempt, which feels quite elegant.

### DIFF
--- a/mijia/src/lib.rs
+++ b/mijia/src/lib.rs
@@ -16,7 +16,7 @@ const SENSOR_READING_CHARACTERISTIC_PATH: &str = "/service0021/char0035";
 const CONNECTION_INTERVAL_CHARACTERISTIC_PATH: &str = "/service0021/char0045";
 /// 500 in little-endian
 const CONNECTION_INTERVAL_500_MS: [u8; 3] = [0xF4, 0x01, 0x00];
-pub const DBUS_METHOD_CALL_TIMEOUT: Duration = Duration::from_secs(30);
+const DBUS_METHOD_CALL_TIMEOUT: Duration = Duration::from_secs(30);
 
 #[derive(Clone, Debug)]
 pub struct SensorProps {
--- a/mijia/src/lib.rs
+++ b/mijia/src/lib.rs
@@ -16,7 +16,7 @@ const SENSOR_READING_CHARACTERISTIC_PATH: &str = "/service0021/char0035";
 const CONNECTION_INTERVAL_CHARACTERISTIC_PATH: &str = "/service0021/char0045";
 /// 500 in little-endian
 const CONNECTION_INTERVAL_500_MS: [u8; 3] = [0xF4, 0x01, 0x00];
-const DBUS_METHOD_CALL_TIMEOUT: Duration = Duration::from_secs(30);
+pub const DBUS_METHOD_CALL_TIMEOUT: Duration = Duration::from_secs(30);
 
 #[derive(Clone, Debug)]
 pub struct SensorProps {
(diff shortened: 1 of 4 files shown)


publish-mqtt/src/main.rs
```diff
@@ -420,42 +422,31 @@ async fn connect_sensor_at_idx(
...
-                        .await
-                        .with_context(|| std::line!().to_string())?;
-                    sensor.connection_status = ConnectionStatus::Disconnected;
-                }
-                ConnectionStatus::Connected => panic!("This should never happen."),
-            };
+
+    let mut backoff = ExponentialBackoff::default();
+    backoff.max_elapsed_time =
+        Some(SENSOR_CONNECT_RESERVATION_TIMEOUT - 2 * DBUS_METHOD_CALL_TIMEOUT);
```
qwandor-google, line 450: I'd rather this just be a constant which is enough small enough than `SENSOR_CONNECT_RESERVATION_TIMEOUT`, rather than making DBUS_METHOD_CALL_TIMEOUT public.

publish-mqtt/src/main.rs
```diff
@@ -420,42 +422,31 @@ async fn connect_sensor_at_idx(
...
+
+    (|| session.start_notify_sensor(id).map_err(Into::into))
+        .retry(backoff)
```
qwandor-google, line 453: Use UFCS for this retry call, as it is confusing for the trait to be added to the closure.

publish-mqtt/src/main.rs
```diff
@@ -220,10 +220,14 @@ impl Sensor {
...
 
 trait SensorStore {
+    fn get_by_id(&self, id: &DeviceId) -> Option<&Sensor>;
     fn get_mut_by_id(&mut self, id: &DeviceId) -> Option<&mut Sensor>;
 }
 
 impl SensorStore for Vec<Sensor> {
+    fn get_by_id(&self, id: &DeviceId) -> Option<&Sensor> {
+        self.iter().find(|s| &s.id == id)
+    }
```
qwandor-google, line 230: Nit: Blank line between methods?

mijia/src/lib.rs
```diff
@@ -16,7 +16,7 @@ const SENSOR_READING_CHARACTERISTIC_PATH: &str = "/service0021/char0035";
 const CONNECTION_INTERVAL_CHARACTERISTIC_PATH: &str = "/service0021/char0045";
 /// 500 in little-endian
 const CONNECTION_INTERVAL_500_MS: [u8; 3] = [0xF4, 0x01, 0x00];
-const DBUS_METHOD_CALL_TIMEOUT: Duration = Duration::from_secs(30);
+pub const DBUS_METHOD_CALL_TIMEOUT: Duration = Duration::from_secs(30);
```
qwandor-google, line 19: I'd prefer this not be public.

publish-mqtt/src/main.rs
```diff
@@ -384,88 +399,77 @@ async fn check_for_sensors(
...
+
     let mut state = state.lock().await;
-    match status {
-        Err(e) => {
-            println!(
-                "Failed to connect to {} (now {:?}): {:?}",
-                sensor.name, sensor.connection_status, e
-            );
-        }
+    let SensorState { sensors, homie, .. } = state.deref_mut();
```
qwandor-google, line 422: This destructuring thing isn't actually needed, you can just use the `let state = &mut *state.lock().await;" trick that we use elsewhere.

qwandor-google (approved):

PR #12, opened 2020-09-26
Generate configs

alsuren:
continuation of https://github.com/alsuren/reading-xiaomi-temp/pull/11

I haven't looked into org.eclipse.smarthome.core.thing.Thing.json properly. If your backups are good then I would be tempted to try running it and see what breaks.

### DIFF
--- a/generate-configs.py
+++ b/generate-configs.py
@@ -98,7 +98,7 @@ def add_named_sensors_to_grafana(mac_to_name, infilename, outfilename):
         f.write("\n")
 
 
-def add_named_sensors_to_smarthome(mac_to_name, infilename, outfilename):
+def add_named_sensors_to_smarthome_items(mac_to_name, infilename, outfilename):
     with open(infilename) as f:
         model = json.load(f)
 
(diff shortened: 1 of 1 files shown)


generate-configs.py
```diff
@@ -98,32 +165,79 @@ def add_named_sensors_to_grafana(mac_to_name, infilename, outfilename):
...
+    smarthome_jsondb_output = os.environ.get("SMARTHOME_JSONDB_OUTPUT")
+    if smarthome_jsondb_input and smarthome_jsondb_output:
+        if not os.path.exists(smarthome_jsondb_output):
+            os.mkdir(smarthome_jsondb_output)
+        add_named_sensors_to_smarthome_items(
+            mac_to_name,
+            f"{smarthome_jsondb_input}/org.eclipse.smarthome.core.items.Item.json",
+            f"{smarthome_jsondb_output}/org.eclipse.smarthome.core.items.Item.json",
+        )
+        # TODO: org.eclipse.smarthome.core.items.Metadata.json
```
qwandor-google, line 230: This is obsolete now, right?

generate-configs.py
```diff
@@ -98,32 +165,79 @@ def add_named_sensors_to_grafana(mac_to_name, infilename, outfilename):
...
+        add_named_sensors_to_smarthome_metadata(
+            mac_to_name,
+            f"{smarthome_jsondb_input}/org.eclipse.smarthome.core.items.Metadata.json",
+            f"{smarthome_jsondb_output}/org.eclipse.smarthome.core.items.Metadata.json",
+        )
+        # TODO: maybe also org.eclipse.smarthome.core.thing.Thing.json?
```
qwandor-google, line 236: I think it is auto-discovered so shouldn't be necessary.

qwandor-google (commented):
This looks reasonable, but it would be good to at least see an example the output it generates, and ideally try it, before submitting.

alsuren:
homie-influx solves the part of this problem that I care about. If we want things added to openhab then the code exists in this branch and can be used whenever.

PR #13, opened 2020-09-28
Iterate over all sensors each time through connection loop

qwandor-google:
Store sensors in a HashMap rather than a Vec to avoid the need for the `SensorStore` methods, and then iterate over them all each time through the connection loop by cloning the set of IDs which key the map.

### DIFF
--- a/publish-mqtt/src/main.rs
+++ b/publish-mqtt/src/main.rs
@@ -10,7 +10,6 @@ use rustls::ClientConfig;
 use std::collections::HashMap;
 use std::fs::File;
 use std::io::{BufRead, BufReader};
-use std::ops::DerefMut;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use tokio::sync::Mutex;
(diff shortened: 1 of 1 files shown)


publish-mqtt/src/main.rs
```diff
@@ -303,34 +287,37 @@ async fn bluetooth_connection_loop(
...
+            for id in ids {
+                let sensor_status = state.lock().await.sensors.get(&id).map(|sensor| {
+                    log::trace!("State of {} is {:?}", sensor.name, sensor.connection_status);
+                    sensor.connection_status
+                });
+                if let Some(connection_status) = sensor_status {
+                    action_sensor(state.clone(), session, id, connection_status)
+                        .await
+                        .with_context(|| std::line!().to_string())?;
+                }
```
alsuren, line 302:
```suggestion
                let connection_status = state.lock().await.sensors.get(&id).map(|sensor| {
                    log::trace!("State of {} is {:?}", sensor.name, sensor.connection_status);
                    sensor.connection_status
                }).expect("sensors cannot be deleted");
                action_sensor(state.clone(), session, id, connection_status)
                    .await
                    .with_context(|| std::line!().to_string())?;
```
seems reasonable to unwrap here

alsuren (approved):

PR #16, opened 2020-10-06
disconnect on watchdog timeout

alsuren:
This is mostly to silence the unused argument warning.

### DIFF
--- a/publish-mqtt/src/main.rs
+++ b/publish-mqtt/src/main.rs
@@ -448,11 +448,18 @@ async fn check_for_stale_sensor(
             sensor.name,
             now - sensor.last_update_timestamp
         );
-        // TODO: Should we disconnect the device first?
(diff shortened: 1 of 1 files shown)


publish-mqtt/src/main.rs
```diff
@@ -448,13 +448,20 @@ async fn check_for_stale_sensor(
...
             .await
             .with_context(|| std::line!().to_string())?;
+        // We could drop our state lock at this point, if it ends up taking
+        // too long. As it is, it's quite nice that we can't attempt to connect
+        // while we're in the middle of disconnecting.
+        session
+            .bt_session
+            .disconnect(&id)
+            .await
+            .with_context(|| std::line!().to_string())?;
```
qwandor-google, line 464: Now that we are disconnecting here, is there any reason to keep the `WatchdogTimeOut` state?

PR #19, opened 2020-10-06
Add homie-controller crate

qwandor-google:


homie-controller/examples/discover.rs
```diff
@@ -0,0 +1,37 @@
...
+                _ => {
+                    println!("Event: {:?}", event);
+                    println!("Devices:");
+                    for device in controller.devices.lock().await.values() {
+                        if device.has_required_attributes() {
+                            println!(" * {:?}", device);
+                        } else {
+                            println!(" * {} not ready.", device.id);
+                        }
+                    }
```
alsuren, line 32: My preference would be for making the current `controller.devices` private and exposing a `controller.for_each_device(|device: & Device| {})` similar to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each (we can't implement Iterator directly, which is a bit annoying. I really wish someone would make a PushIterator/InternalIterator trait that we could use in times like these)

something like:
```suggestion
                    controller.for_each_device(|device| {
                        if device.has_required_attributes() {
                            println!(" * {:?}", device);
                        } else {
                            println!(" * {} not ready.", device.id);
                        }
                    })
```

When we come across use-cases that require other ways to access the device tree then I'm happy to iterate on it, but handing out mutable access to our internal state feels like the wrong answer.

In general, I would be in favour of iterating on our API design by building tools that solve real problems (I've been thinking about http://hintjens.com/blog:106 a bunch recently, and pondering how I can get closer to that ideal).

homie-controller/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+            mqtt_client,
+            base_topic: base_topic.to_string(),
+            devices: Arc::new(Mutex::new(HashMap::new())),
+        };
+        (controller, event_loop)
+    }
+
+    /// Poll the `EventLoop`, and maybe return a Homie event.
+    pub async fn poll(&self, event_loop: &mut EventLoop) -> Result<Option<Event>, PollError> {
+        let notification = event_loop.poll().await?;
```
alsuren, line 120: I think the problem is that we're handing out event_loop in a way that our api consumer can decide whether they poll it directly or poll it via this method.

Could we wrap it in an opaque type that has a poll method?

It could be the simple reverse of this function signature (`HomieEventLoop::poll(&mut self, &HomieController)`) or HomieEventLoop could hold a reference to HomieController somehow, and make it `HomieEventLoop::poll(&mut self)`

homie-controller/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+    }
+}
+
+/// A Homie controller, which connects to an MQTT broker and interacts with Homie devices.
+#[derive(Debug)]
+pub struct HomieController {
+    mqtt_client: AsyncClient,
+    base_topic: String,
+    /// The set of Homie devices which have been discovered so far, keyed by their IDs.
+    pub devices: Arc<Mutex<HashMap<String, Device>>>,
```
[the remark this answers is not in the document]
qwandor-google: As far as I can see, `im::HashMap<...>` has the same semantics as `Arc<HashMap<...>>`, though potentially more efficient in the case that it does need to make a copy. The downside is that it is an extra dependency, and a different type in the API. Changed to the latter for now, pending further thought.

homie-controller/src/lib.rs
qwandor-google: I can't figure out how to make `HomieEventLoop` include a reference to `HomieController`, unless we always wrap `HomieController` in an `Arc`. Maybe we can talk this through some more.

alsuren: I can't see a way around Arc either. Maybe it's something to think about in another pr.

homie-controller/src/lib.rs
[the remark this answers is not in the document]
alsuren: Yeah. Seems good enough. Would be interesting to see some unit tests, potentially using https://doc.rust-lang.org/1.30.0/std/ptr/fn.eq.html to catch when things get cloned.

alsuren (approved):
I think this looks promising. Let's merge it and iterate on the fiddly bits once we have more experience from building things on to of it.

homie-controller/examples/discover.rs
```diff
@@ -0,0 +1,37 @@
...
+use rumqttc::MqttOptions;
+
+#[tokio::main(core_threads = 2)]
+async fn main() -> Result<(), PollError> {
+    pretty_env_logger::init();
+
+    let mqttoptions = MqttOptions::new("homie_controller", "test.mosquitto.org", 1883);
+
+    let (controller, mut event_loop) = HomieController::new(mqttoptions, "homie");
+    controller.start().await?;
```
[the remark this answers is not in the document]
qwandor-google: I'd rather keep starting separate from polling, in case the user of the library wants to set up a thread to poll before calling `start()`.

homie-controller/examples/discover.rs
qwandor-google: I'd like to make it easy for users of the library not just to iterate over devices but also to look up specific ones. I agree that we shouldn't be handing out mutable access to them though, that is at least fixed now.

homie-controller/src/lib.rs
qwandor-google: Wrapped in an opaque type for now. I'll consider an Arc on `HomieEventLoop` or a Mutex on `HomieController` later.

PR #23, opened 2020-10-08
Make devices() not async, and unsubscribe from removed properties and nodes

qwandor-google:


homie-controller/src/lib.rs
```diff
@@ -170,7 +169,34 @@ impl HomieController {
...
+
+        Ok(event)
+    }
+
+    /// Handle a publish event, update the devices, and return any event and any new topics which
+    /// should be subscribed to or unsubscribed from.
+    ///
+    /// This is separate from `handle_publish` because it takes the `devices` lock, to ensure that
+    /// no async operations are awaited while the lock is held.
+    fn handle_publish_update_devices(
```
qwandor-google, line 196: This method needs a better name.

alsuren: I would be okay with handle_publish_sync or update_devices. It's a bit fishy that it mutates things and also returns something. I suppose it could receive devices by ownership and return devices by ownership, along with the other arguments and return types, and then the name would be less important, because the arguments would tell you what it's doing.

alsuren (approved):
The behaviour looks right. There are some refactors lurking in the shadows, but I'm not sure what they are. Happy to get it merged and think about it later, since the user facing changes are definitely worthwhile.

homie-controller/src/lib.rs
```diff
@@ -182,67 +208,107 @@ impl HomieController {
...
-                return Ok(Some(Event::device_updated(device)));
+                Some(Event::device_updated(device))
             }
             [device_id, "$nodes"] => {
                 let nodes: Vec<_> = payload.split(",").collect();
                 let device = get_mut_device_for(devices, "Got nodes for", device_id)?;
+
                 // Remove nodes which aren't in the new list.
-                device.nodes.retain(|k, _| nodes.contains(&k.as_ref()));
+                device.nodes.retain(|node_id, node| {
```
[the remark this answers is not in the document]
alsuren: Ah. The pattern I was searching for was react's virtual dom.

If you have a device tree that's updated by these events, and a subscription tree that's managed by a separate process [the subscription tree might be a flat hashmap, but I will keep calling it a tree for now, because editing on my phone is hard], then you can calculate the desired subscription tree from the device tree, and then take a diff to determine which subscriptions to add or remove.

I'm assuming that there is a handshake for subscribing that goes 
```
subscribe ->
<- subscribed
```
or something. If that is the case then your subscription tree wants to have leaves like SubscriptionRequested (ts), Subscribed, Unsubscription requested, [Unsubscribed]

Where Unsubscribed might be represented by removing the node from the tree.

This would allow better separation of concerns, and simplify event handling. It would also make the topic subscription state machine explicit and observable when debugging is needed.

homie-controller/src/lib.rs
qwandor-google: Yeah, the split is really just because it does the parts of `handle_publish` which don't need to call `.await`. It could be a block within `handle_publish`, except that it would be too easy for someone to later accidentally add a `.await` in where it shouldn't be.

homie-controller/src/lib.rs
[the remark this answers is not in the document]
qwandor-google: I guess there is something like this handshake happening, but we are currently not looking at that level; we just assume that if we send a subscription request then the MQTT client library ensures that the server gets it and responds to it. Keeping track of that state here seems like making unnecessary work for ourselves which should already be handled at a lower layer.

Issue #21, opened 2020-10-07
replace read-all-devices.


alsuren:
Its different use-cases should be split into:
- [ ] a no-mqtt mode for publish-mqtt.
- [ ] an example in mijia crate that takes a bluetooth address as env var.
- [ ] an example in mijia crate that continuously scans and connects to each sensor in turn for a configurable time/reading-count.

qwandor:
Discussed in person, the first two use-cases listed above are essentially covered by the `readings`. Something that handles retries and disconnecting might be useful for the third, for handling more sensors than the Bluetooth adapter supports simultaneous connections.

PR #25, opened 2020-10-10
Mijia example and logging

qwandor-google:
Add an example of discovering Mijia sensors, and some more logging including device names.

### DIFF
--- a/mijia/Cargo.toml
+++ b/mijia/Cargo.toml
@@ -13,4 +13,8 @@ dbus-tokio = "0.5.2"
 eyre = "0.6"
 futures = "0.3.5"
 itertools = "0.9.0"
+log = "0.4.11"
 tokio = "0.2.22"
+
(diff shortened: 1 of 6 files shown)


mijia/Cargo.toml
```diff
@@ -13,4 +13,6 @@ dbus-tokio = "0.5.2"
 eyre = "0.6"
 futures = "0.3.5"
 itertools = "0.9.0"
+log = "0.4.11"
+pretty_env_logger = "0.4.0"
```
alsuren, line 17: Should these be dev dependencies if they're only used by examples?

qwandor-google: `log` is needed for the library, but `pretty_env_logger` you're right.

qwandor-google:
This helps a bit with #21 too.

alsuren (approved):

PR #26, opened 2020-10-12
homie-device: Use new AsyncClient.

qwandor-google:


homie-device/src/lib.rs
```diff
@@ -489,29 +488,23 @@ impl DevicePublisher {
...
-        Ok(())
+        self.client
+            .publish(topic, QoS::AtLeastOnce, true, value)
+            .await
     }
 
     async fn disconnect(&self) -> Result<(), ClientError> {
-        self.requests_tx.send(Request::Disconnect).await?;
-        Ok(())
+        self.client.disconnect().await
```
alsuren, line 497: Maybe just delete this method and use the one on client instead?

publish-mqtt/src/main.rs
```diff
@@ -1,3 +1,5 @@
+#![type_length_limit = "1129369"]
```
alsuren, line 1: :-(

alsuren (approved):
This is a lot easier to read than before.

homie-device/src/lib.rs
qwandor-google: Done.

PR #27, opened 2020-10-12
Filter Mijia sensors on name rather than service data UUID

qwandor-google:


mijia/src/bluetooth.rs
```diff
@@ -220,3 +196,38 @@ impl BluetoothSession {
...
+
+fn get_service_data(
+    device_properties: &HashMap<String, Variant<Box<dyn RefArg>>>,
+) -> HashMap<String, Vec<u8>> {
+    // UUIDs don't get populated until we connect. Use:
+    //     "ServiceData": Variant(InternalDict { data: [
+    //         ("0000fe95-0000-1000-8000-00805f9b34fb", Variant([48, 88, 91, 5, 1, 23, 33, 215, 56, 193, 164, 40, 1, 0])
+    //     )], outer_sig: Signature("a{sv}") })
+    // instead.
+    || -> Option<HashMap<_, _>> {
```
alsuren, line 208: make this function return Option and make get_devices() use unwrap_or_default() on it.

alsuren (approved):
seems to work on the device. Tiny refactor then merge it.

mijia/src/bluetooth.rs
qwandor-google: Done.

PR #29, opened 2020-10-13
Implement legacy firmware and legacy stats extensions for Homie controller

qwandor-google:


homie-controller/src/lib.rs
```diff
@@ -248,6 +252,82 @@ impl HomieController {
...
+                let device = get_mut_device_for(devices, "Got localip for", device_id)?;
+                device.local_ip = Some(payload.to_owned());
+                Some(Event::device_updated(device))
+            }
+            [device_id, "$mac"] => {
+                let device = get_mut_device_for(devices, "Got mac for", device_id)?;
+                device.mac = Some(payload.to_owned());
+                Some(Event::device_updated(device))
+            }
+            [device_id, "$fw", "name"] => {
```
alsuren, line 273: Might be clearer to use `let parts = subtopic.splitn("/", 2).collect::<Vec<&str>>();` above, and `[device_id, "$fw/name"]` here, now that this is the majority case.

I would also start thinking about using a macro at this point. Would it be possible to write this as:

```
event_case!("$fw/name", firmware_name, std::borrow::ToOwned::to_owned)
```
 or something? (it might not be possible. I have no idea).

homie-controller/src/types.rs
```diff
@@ -245,6 +278,45 @@ pub struct Device {
 
     /// The nodes of the device, keyed by their IDs.
     pub nodes: HashMap<String, Node>,
+
+    /// The Homie extensions implemented by the device.
+    pub extensions: Vec<Extension>,
+
+    /// The IP address of the device on the local network.
+    pub local_ip: Option<String>,
```
alsuren, line 286: How big is the homie spec? Is it worth having a `HashMap<String, String>` of overflow properties?

I suppose that might be hard to do in a backwards-compatible way, because as soon as you add first-class support for something, it stops existing in the overflow properties.

alsuren: ... also: how big is this struct in bytes? How big and sparse does it need to get before it would be more efficient to store everything in a hashmap and expose getters for each of the properties? ("we don't expect to ever handle enough devices to care" is a valid answer to these questions)

homie-controller/src/lib.rs
```diff
@@ -5,15 +5,18 @@ use rumqttc::{
...
 use thiserror::Error;
 use tokio::task::JoinError;
 
 mod types;
-pub use types::{Datatype, Device, Node, ParseDatatypeError, ParseStateError, Property, State};
+pub use types::{Datatype, Device, Extension, Node, Property, State};
+use types::{ParseDatatypeError, ParseExtensionError, ParseStateError};
 
-const REQUESTS_CAP: usize = 10;
+const REQUESTS_CAP: usize = 100;
```
alsuren, line 19: Is this just pushing the deadlock into the middle-distance, or there something specific about the deadlock case that means that it is solved completely at 100? I guess what I mean is: could we push it into the far-distance by increasing to 1000?

homie-controller/src/types.rs
```diff
@@ -219,6 +220,38 @@ impl Node {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let parts: Vec<_> = s.split(":").collect();
+        if let [id, version, homie_versions] = parts.as_slice() {
+            if let Some(homie_versions) = homie_versions.strip_prefix("[") {
+                if let Some(homie_versions) = homie_versions.strip_suffix("]") {
+                    return Ok(Extension {
+                        id: (*id).to_owned(),
+                        version: (*version).to_owned(),
+                        homie_versions: homie_versions.split(";").map(|p| p.to_owned()).collect(),
```
alsuren, line 246: It feels like this wants a couple of tests.

homie-controller/src/lib.rs
```diff
@@ -248,6 +252,82 @@ impl HomieController {
...
+                device.local_ip = Some(payload.to_owned());
+                Some(Event::device_updated(device))
+            }
+            [device_id, "$mac"] => {
+                let device = get_mut_device_for(devices, "Got mac for", device_id)?;
+                device.mac = Some(payload.to_owned());
+                Some(Event::device_updated(device))
+            }
+            [device_id, "$fw", "name"] => {
+                let device = get_mut_device_for(devices, "Got fw/name for", device_id)?;
```
alsuren, line 274: thinking about this further, I'd be slightly more happy with something of the form `let device = get_mut_device_for_update(devices, device_id, "$fw/name")?;`

If we go down the macro route then we could even do the error message formatting in the macro and kill off get_mut_device_for() entirely.

alsuren (approved):
Code looks good. Entirely optional comments, and suggestions for future direction.

homie-controller/src/lib.rs
qwandor-google: How would that work for the node and property attributes below?

homie-controller/src/types.rs
qwandor-google: I would prefer them to be typed, though some way to separate extensions (like the ones added in this PR) from the core spec might be good. I'm not sure how best to do that though, or whether it's actually worthwhile.

homie-controller/src/lib.rs
qwandor-google: I was thinking of doing that originally, but then it doesn't work quite so well calling it from `get_mut_node_for`.

homie-controller/src/types.rs
alsuren: Might be interesting to separate the extensions out with traits. Not needed for the current feature set though, so we have some time to ponder how to do it.

homie-controller/src/types.rs
qwandor-google: Added.

homie-controller/src/lib.rs
qwandor-google: Done.

PR #31, opened 2020-10-17
Add method on Controller to set a property value, and an example of how to use it

qwandor-google:
The example turns all devices on and off repeatedly.

### DIFF
--- a/homie-controller/examples/flash.rs
+++ b/homie-controller/examples/flash.rs
@@ -63,6 +63,8 @@ async fn main() -> Result<(), PollError> {
         }
     }
 
+    controller.disconnect().await?;
+
     try_join!(handle.map(|res| Ok::<_, PollError>(res??)))?;
 
     Ok(())
--- a/homie-controller/examples/flash.rs
+++ b/homie-controller/examples/flash.rs
@@ -1,0 +1,69 @@
+//! Example to turn all Homie devices on and off every 5 seconds, 4 times.
+
+use futures::FutureExt;
+use homie_controller::{Datatype, Event, HomieController, HomieEventLoop, PollError};
+use rumqttc::MqttOptions;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::task::{self, JoinHandle};
+use tokio::{time, try_join};
+
+fn spawn_poll_loop(
+    mut event_loop: HomieEventLoop,
+    controller: Arc<HomieController>,
+) -> JoinHandle<Result<(), PollError>> {
+    task::spawn(async move {
+        loop {
+            if let Some(event) = controller.poll(&mut event_loop).await? {
+                match event {
+                    Event::PropertyValueChanged {
+                        device_id,
+                        node_id,
+                        property_id,
+                        value,
+                    } => {
+                        println!("{}/{}/{} = {}", device_id, node_id, property_id, value);
+                    }
+                    _ => {
+                        log::info!("Event: {:?}", event);
+                    }
+                }
+            }
+        }
+    })
+}
+
+#[tokio::main]
+async fn main() -> Result<(), PollError> {
+    pretty_env_logger::init();
+
+    let mqttoptions = MqttOptions::new("homie_controller", "test.mosquitto.org", 1883);
+
+    let (controller, event_loop) = HomieController::new(mqttoptions, "homie");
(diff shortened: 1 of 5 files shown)


homie-controller/examples/flash.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let mqttoptions = MqttOptions::new("homie_controller", "test.mosquitto.org", 1883);
+
+    let (controller, event_loop) = HomieController::new(mqttoptions, "homie");
+    let controller = Arc::new(controller);
+    let handle = spawn_poll_loop(event_loop, controller.clone());
+    controller.start().await?;
+
+    for _ in 0..3 {
+        for &value in [true, false].iter() {
```
alsuren, line 48:
```suggestion
        for value in [true, false].into_iter() {
```

Not sure if array even has that method

homie-controller/examples/flash.rs
```diff
@@ -0,0 +1,74 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    controller.disconnect().await?;
+
+    try_join!(handle.map(|res| {
+        let r: Result<_, PollError> = Ok(res??);
```
alsuren, line 69:
```suggestion
        Ok::<_, PollError>(res??)
```

And then remove the rest of the closure and fmt.
It may be that cargo clippy has something to say about this.

homie-controller/examples/flash.rs
```diff
@@ -0,0 +1,74 @@
...
+        for &value in [true, false].iter() {
+            time::delay_for(Duration::from_secs(5)).await;
+            println!("Turning everything {}", if value { "on" } else { "off" });
+            for device in controller.devices().values() {
+                for node in device.nodes.values() {
+                    for property in node.properties.values() {
+                        if property.settable && property.datatype == Some(Datatype::Boolean) {
+                            println!("{}/{}/{} set to {}", device.id, node.id, property.id, value);
+                            controller
+                                .set(&device.id, &node.id, &property.id, value)
```
alsuren, line 57: .set() feels like the lowest-level-possible implementation of the thing. Are there plans for a wrapper that does the settableness check and waits for the property to change in the remote end? (May not be possible to do this)

homie-controller/examples/flash.rs
```diff
@@ -0,0 +1,74 @@
...
+use rumqttc::MqttOptions;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::task::{self, JoinHandle};
+use tokio::{time, try_join};
+
+fn spawn_poll_loop(
+    mut event_loop: HomieEventLoop,
+    controller: Arc<HomieController>,
+) -> JoinHandle<Result<(), PollError>> {
```
alsuren, line 14: This definitely wants to live in the library in some form. Let's wait until we have a few more examples and then sit down and work out what form it should take.

homie-controller/examples/flash.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let (controller, event_loop) = HomieController::new(mqttoptions, "homie");
+    let controller = Arc::new(controller);
+    let handle = spawn_poll_loop(event_loop, controller.clone());
+    controller.start().await?;
+
+    for _ in 0..3 {
+        for &value in [true, false].iter() {
+            time::delay_for(Duration::from_secs(5)).await;
+            println!("Turning everything {}", if value { "on" } else { "off" });
```
alsuren, line 50: Oh wow. I feel like I need to try this out now.

alsuren (approved):
Minor comments.

Can I request something in the description field (even if it's just a pointless gif)? I feel sad seeing it empty.

homie-controller/examples/flash.rs
qwandor-google: That doesn't seem to work.

homie-controller/examples/flash.rs
qwandor-google: Aha! I was trying to figure out the syntax for this and somehow missed the obvious.

homie-controller/examples/flash.rs
qwandor-google: Potentially. I haven't figured out how this should look yet. Waiting might not be a good idea, as it might never be updated if a device is misbehaving.

Issue #35, opened 2020-10-17
Improve handling of Homie typed properties


alsuren:

qwandor:
`homie-controller` is looking pretty good now, but it might be good to do something similar for `homie-device`.

PR #38, opened 2020-10-17
Add getters to parse property value and format

qwandor-google:
This helps towards #35 for `homie-controller`.

### DIFF
--- a/homie-controller/src/types.rs
+++ b/homie-controller/src/types.rs
@@ -234,7 +234,15 @@ impl Property {
 
         match self.format {
             None => Err(ValueError::Unknown),
-            Some(ref format) => Ok(format.split(',').collect()),
+            Some(ref format) => {
+                if format.is_empty() {
+                    Err(ValueError::WrongFormat {
+                        format: "".to_owned(),
+                    })
+                } else {
+                    Ok(format.split(',').collect())
+                }
+            }
         }
     }
 }
@@ -677,9 +685,14 @@ mod tests {
         // With no known format or datatype, format parsing fails.
         assert_eq!(property.enum_values(), Err(ValueError::Unknown));
 
-        // An empty string is possible.
+        // An empty format string is invalid.
         property.format = Some("".to_owned());
-        assert_eq!(property.enum_values(), Ok(vec![]));
+        assert_eq!(
+            property.enum_values(),
+            Err(ValueError::WrongFormat {
+                format: "".to_owned()
+            })
+        );
 
         // A single value is valid.
         property.format = Some("one".to_owned());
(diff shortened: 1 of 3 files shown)


qwandor-google:
TODO:
- [ ] Parse range format for integer and float datatype.

homie-controller/src/types.rs
```diff
@@ -394,4 +478,283 @@ mod tests {
...
+            })
+        );
+
+        // With a valid value but unknown datatype, parsing succeeds.
+        property.value = Some("42".to_owned());
+        assert_eq!(property.value(), Ok(42));
+
+        // With the correct datatype, parsing still succeeds.
+        property.datatype = Some(Datatype::Integer);
+        assert_eq!(property.value(), Ok(42));
```
alsuren, line 505: I have a strong preference for one test per function, so that you can work out what's going on when you make a bunch of them fail. I would suggest:

* Extract each of the comments to be test names.
* potentially use nested `mod`s to help with grouping of tests
* There are a few ways to make instantiating structs easier (make a `prop("42")` helper, and do `Property { datatype: Some(Datatype::Integer), ..prop("42") }`

https://crates.io/crates/test-case will also help, if you want something that's more like table driven testing.

alsuren: ... that said, I can hardly dictate the style of tests in the project if I refuse to write any, so maybe merge it and I'll show you what I mean when I finally write some tests.

homie-controller/src/types.rs
```diff
@@ -167,6 +187,64 @@ impl Property {
     pub fn has_required_attributes(&self) -> bool {
         self.name.is_some() && self.datatype.is_some()
     }
+
+    pub fn value<T: Value>(&self) -> Result<T, ValueError> {
```
alsuren, line 191: I'm a little bit uncomfortable with having a function that has the same name as a property. Looks like the compiler is able to work out what you mean though, so maybe it's fine?

qwandor-google: What would you suggest calling it instead, is there some typical naming convention?

homie-controller/src/types.rs
qwandor-google: You mean one test per assert? That seems a bit over the top I think. In this case I quiet like modifying the property and checking that I get the results I expect each step of the way.

alsuren: I found it a bit difficult to follow which of your mutations were still in flight and interesting vs which had been shadowed vs which were in flight but not interesting to the current assertion.

Splitting tests doesn't have to be much more verbose if you have easy ways to construct the various states, and it can be much easier to follow and debug.

A comment thread is probably not the best place to discuss this. Let's either pair on it, or merge it and I will have a go at refactoring at a later date.

homie-controller/src/types.rs
alsuren: Honestly not sure. I will ask at work. Might be that your approach is how everyone does it.

alsuren: Ah. Apparently it is common, but normally the field is private.

qwandor-google: Okay, I'll keep it as is.

PR #43, opened 2020-10-19
Add more tests for homie-controller

qwandor-google:
Better late than never.

### DIFF
--- a/homie-controller/src/lib.rs
+++ b/homie-controller/src/lib.rs
@@ -617,3 +617,212 @@ impl From<ParseFloatError> for HandleError {
         HandleError::Warning(format!("Invalid float: {}", e))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use async_channel::Receiver;
+    use rumqttc::{Packet, Request, Subscribe};
+
+    fn make_test_controller() -> (HomieController, Receiver<Request>) {
+        let (requests_tx, requests_rx) = async_channel::unbounded();
+        let (cancel_tx, _cancel_rx) = async_channel::unbounded();
+        let mqtt_client = AsyncClient::from_senders(requests_tx, cancel_tx);
+        let controller = HomieController {
+            base_topic: "base_topic".to_owned(),
+            mqtt_client,
+            devices: Mutex::new(Arc::new(HashMap::new())),
+        };
+        (controller, requests_rx)
+    }
+
+    fn expect_subscribe(request: Request, expected: Subscribe) {
+        if let Request::Subscribe(actual) = request {
+            assert_eq!(actual, expected);
+        } else {
+            panic!("Expected subscribe but got {:?}", request);
+        }
+    }
+
+    async fn publish(
+        controller: &HomieController,
+        topic: &str,
+        payload: &str,
+    ) -> Result<Option<Event>, PollError> {
+        controller
+            .handle_event(Packet::Publish(Publish::new(
+                topic,
+                QoS::AtLeastOnce,
+                payload,
+            )))
+            .await
+    }
+
+    #[tokio::test]
+    async fn controller_start() -> Result<(), Box<dyn std::error::Error>> {
+        let (controller, requests_rx) = make_test_controller();
+
+        // Start discovering.
+        controller.start().await?;
+        expect_subscribe(
+            requests_rx.recv().await?,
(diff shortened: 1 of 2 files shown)


homie-controller/src/lib.rs
```diff
@@ -617,3 +616,206 @@ impl From<ParseFloatError> for HandleError {
...
+            Subscribe::new("base_topic/device_id/+", QoS::AtLeastOnce),
+        );
+        expect_subscribe(
+            requests_rx.recv().await?,
+            Subscribe::new("base_topic/device_id/$fw/+", QoS::AtLeastOnce),
+        );
+        expect_subscribe(
+            requests_rx.recv().await?,
+            Subscribe::new("base_topic/device_id/$stats/+", QoS::AtLeastOnce),
+        );
```
alsuren, line 690: (crazy idea): Could we have something like this that doesn't care about ordering?
```suggestion
        expect_subscriptions(
            requests_rx,
            vec!["base_topic/device_id/+", "base_topic/device_id/$fw/+", "base_topic/device_id/$stats/+"],
        );
```

homie-controller/src/lib.rs
```diff
@@ -620,3 +619,206 @@ impl From<ParseFloatError> for HandleError {
...
+            base_topic: "base_topic".to_owned(),
+            mqtt_client,
+            devices: Mutex::new(Arc::new(HashMap::new())),
+        };
+        (controller, requests_rx)
+    }
+
+    fn expect_subscribe(request: Request, expected: Subscribe) {
+        if let Request::Subscribe(actual) = request {
+            assert_eq!(actual, expected);
```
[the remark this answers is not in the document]
alsuren: Also, it looks like we always subscribe with QoS::AtLeastOnce in tests, so this helper could just hard-code it.

homie-controller/src/types.rs
```diff
@@ -824,4 +840,63 @@ mod tests {
...
+        assert_eq!(node.has_required_attributes(), false);
+
+        node.name = Some("Node name".to_owned());
+        assert_eq!(node.has_required_attributes(), false);
+
+        node.node_type = Some("Node type".to_owned());
+        assert_eq!(node.has_required_attributes(), false);
+
+        node.add_property(property_with_required_attributes());
+        assert_eq!(node.has_required_attributes(), true);
```
alsuren, line 876: so node needs at least one property? I suppose that makes sense.

homie-controller/src/types.rs
```diff
@@ -824,4 +840,63 @@ mod tests {
...
+        assert_eq!(device.has_required_attributes(), false);
+
+        device.name = Some("Device name".to_owned());
+        assert_eq!(device.has_required_attributes(), false);
+
+        device.state = State::Init;
+        assert_eq!(device.has_required_attributes(), true);
+
+        device.add_node(node_with_required_attributes());
+        assert_eq!(device.has_required_attributes(), true);
```
alsuren, line 900: if you add a node without the required attributes, does it go to false?

homie-controller/src/lib.rs
```diff
@@ -617,3 +616,206 @@ impl From<ParseFloatError> for HandleError {
...
+                "Node type"
+            )
+            .await?,
+            Some(Event::NodeUpdated {
+                device_id: "device_id".to_owned(),
+                node_id: "node_id".to_owned(),
+                has_required_attributes: false
+            })
+        );
+        let mut expected_node = Node::new("node_id");
```
alsuren, line 751: I still don't like these variables that evolve throughout the test. Let me try something.

homie-controller/src/lib.rs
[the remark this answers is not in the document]
qwandor: 'Request' unfortunately does not currently derive `PartialEq`, which is why I added this helper function. I sent https://github.com/bytebeamio/rumqtt/pull/171 to fix that; once that gets released we can clean this up.

homie-controller/src/lib.rs
qwandor: Good idea, I'll do that.

homie-controller/src/types.rs
qwandor: That's the current logic, yes. Whether that's how it should be I'm not entirely sure, and I'm inconsistent with devices.

homie-controller/src/types.rs
qwandor: Added.

PR #48, opened 2020-10-21
Controller docs

qwandor:
More docs for `homie-controller`.

### DIFF
--- a/homie-controller/src/types.rs
+++ b/homie-controller/src/types.rs
@@ -75,11 +75,19 @@ impl Display for State {
 /// The data type of a Homie property.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum Datatype {
+    /// A [64-bit signed integer](https://homieiot.github.io/specification/#integer).
     Integer,
+    /// A [64-bit floating-point number](https://homieiot.github.io/specification/#float).
     Float,
+    /// A [boolean value](https://homieiot.github.io/specification/#boolean).
     Boolean,
+    /// A [UTF-8 encoded string](https://homieiot.github.io/specification/#string).
     String,
+    /// An [enum value](https://homieiot.github.io/specification/#enum) from a set of possible
+    /// values specified by the property format.
     Enum,
+    /// An RGB or HSV [color](https://homieiot.github.io/specification/#color), depending on the
+    /// property format.
     Color,
 }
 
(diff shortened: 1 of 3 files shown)


homie-controller/src/types.rs
```diff
@@ -75,11 +75,19 @@ impl Display for State {
...
     Float,
+    /// A [boolean value](https://homieiot.github.io/specification/#boolean).
     Boolean,
+    /// A [UTF-8 encoded string](https://homieiot.github.io/specification/#string).
     String,
+    /// An [enum value](https://homieiot.github.io/specification/#enum) from a set of possible
+    /// values specified by the property format.
     Enum,
+    /// An RGB or HSV [color](https://homieiot.github.io/specification/#color), depending on the
+    /// property format.
```
alsuren, line 90: (optional) can this link to `Property::format`?

homie-controller/src/types.rs
```diff
@@ -157,9 +165,16 @@ pub struct Property {
 
     /// The format of the property, if any. This should be specified if the datatype is `Enum` or
     /// `Color`, and may be specified if the datatype is `Integer` or `Float`.
+    ///
+    /// This field holds the raw string received from the device; use
```
alsuren, line 169: ; is distracting
```suggestion
    /// This field holds the raw string received from the device. Use
```
(same below)

qwandor: Done.

homie-controller/src/types.rs
qwandor: Linked to Property::color_format() instead.

homie-controller/src/values.rs
```diff
@@ -4,19 +4,36 @@ use std::num::ParseIntError;
...
+    /// by the device was something different.
     #[error("Expected value of type {expected} but was {actual}.")]
     WrongDatatype {
+        /// The datatype expected by the method call.
         expected: Datatype,
+        /// The actual datatype of the property, as sent by the device.
         actual: Datatype,
     },
+    /// The format of the property couldn't be parsed or didn't match what was expected by the
+    /// method call.
```
alsuren, line 23: (for another day): is there any way to split these two cases, and/or show "what was expected" to help with debugging?

alsuren (approved):

homie-controller/src/values.rs
qwandor: I originally had two different enum variants, but I'm not convinced it's a useful distinction.

As for "what was expected", there's not necessarily an exact expected value. If you call `value::<ColorRGB>` we expect the string "rgb", but if you call `enum_values` or `range` then we just expect it to follow the right format.

Issue #18, opened 2020-10-06
Build Debian package for release


qwandor-google:
It would be useful to build Debian packages for Raspberry Pi (and maybe other targets) that users can install rather than building from source.

qwandor:
Once cargo-deb is sorted, https://www.reddit.com/r/rust/comments/ht0p25/github_actions_for_packaging_deb_rpm_statically/ or https://github.com/NLnetLabs/krill/blob/master/.github/workflows/pkg.yml#L56 may be useful for figuring out how to get cross-compilation working with Github Actions.

qwandor:
Next steps: trigger the package workflow when mijia-home-* tag is pushed, if the packaging is successful then [create a draft release](https://github.com/marketplace/actions/create-a-release) and add the Debian packages to it. When release is (manually) published, push them to the repository.

qwandor:
Remaining things to do:
 - [ ] Generate a suitable GPG keypair and enable signing on https://bintray.com/alsuren/mijia-homie.
 - [ ] Fill in the description and other details on https://bintray.com/alsuren/mijia-homie.

qwandor:
Moved to https://bintray.com/homie-rs/homie-rs instead, and filled in details there.

PR #55, opened 2020-10-24
Build Debian packages on push to master

qwandor:
Also a simple script to do it locally.

This has to work around https://github.com/mmstick/cargo-deb/issues/151, which is a bit annoying.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -69,6 +69,14 @@ jobs:
     - name: Install dependencies
       run: sudo apt-get install libdbus-1-dev
 
+    - name: Cache .cargo and target
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/git
+          ~/.cargo/registry
+          target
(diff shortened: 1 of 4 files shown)


.github/workflows/package.yml
```diff
@@ -0,0 +1,92 @@
...
+      id: cache-cargo-deb
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/bin/cargo-deb
+          ~/.cargo/bin/cross
+          ~/.cargo/.crates.toml
+          ~/.cargo/.crates2.json
+        key: ${{ runner.os }}-cargo-deb-${{ env.cargo-deb-version }}
+
```
alsuren, line 82: cache target as well?

alsuren (approved):
- [ ] merge and then check that the debian package works on the raspberry pis
- [x] Make a task for the auto-dep warning

.github/workflows/package.yml
qwandor: Whoops, yes.

qwandor:
This works towards #18.

PR #60, opened 2020-10-26
Add aarch64 support back, and build in Package workflow too.

qwandor:
More work towards #18.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -9,21 +9,28 @@ env:
   cargo-deb-version: 1.28.0
 
 jobs:
-  build-armv7:
+  build-all:
     runs-on: ubuntu-latest
-    env:
-      target: armv7-unknown-linux-gnueabihf
-      binutils: binutils-arm-linux-gnueabihf
+    strategy:
+      matrix:
+        target:
+          - armv7-unknown-linux-gnueabihf
+          - aarch64-unknown-linux-gnu
+        include:
+          - target: armv7-unknown-linux-gnueabihf
+            binutils: binutils-arm-linux-gnueabihf
+          - target: aarch64-unknown-linux-gnu
+            binutils: binutils-aarch64-linux-gnu
 
     steps:
       - uses: actions/checkout@v2
       - name: Install dependencies
-        run: sudo apt-get install libdbus-1-dev ${{ env.binutils }}
+        run: sudo apt-get install libdbus-1-dev ${{ matrix.binutils }}
       - name: Toolchain
         uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-          target: ${{ env.target }}
+          target: ${{ matrix.target }}
           override: true
 
       - name: Cache .cargo and target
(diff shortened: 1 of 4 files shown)


.github/workflows/package.yml
```diff
@@ -2,99 +2,79 @@ name: Package
...
-    - name: Install cargo-deb
-      run: cargo install cargo-deb --version ${{ env.cargo-deb-version }}
+      - name: Cache .cargo and target
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/git
+            ~/.cargo/registry
+            target
+          key: ${{ runner.os }}-package-cargo-${{ hashFiles('**/Cargo.lock') }}
```
alsuren, line 45: also matrix.target to reduce the download size? (How big is it before/after this change, anyway?)

.github/workflows/package.yml
```diff
@@ -2,99 +2,79 @@ name: Package
...
-        path: |
-          ~/.cargo/bin/cargo-deb
-          ~/.cargo/bin/cross
-          ~/.cargo/.crates.toml
-          ~/.cargo/.crates2.json
-        key: ${{ runner.os }}-cargo-deb-${{ env.cargo-deb-version }}
+      # If a target is set, build with cross.
+      - name: Cross build
+        if: matrix.target != ''
+        uses: actions-rs/cargo@v1
```
alsuren, line 62: If I disable my docker daemon and run `cross build` with no arguments or `cross build --target x86_64-apple-darwin` then it Just Works (the former puts stuff in ./target/debug and the latter in ./target/x86_64-apple-darwin/debug)

I think that our build scripts would be greatly simplified if we use cross for all targets in the matrix. It may take longer to install the extra toolchain, but I think that's fine.

run.sh
```diff
@@ -16,7 +16,7 @@ EXAMPLE=${EXAMPLE:-}
 # Target architecture for Raspbian on a Raspberry Pi.
 # Changing this requires changes to Cross.toml. Send a patch if you want this
 # to be made configurable again.
```
alsuren, line 18: comment probably needs updating

.github/workflows/package.yml
```diff
@@ -2,99 +2,79 @@ name: Package
...
     runs-on: ubuntu-latest
-    env:
-      target: armv7-unknown-linux-gnueabihf
-      binutils: binutils-arm-linux-gnueabihf
+    strategy:
+      matrix:
+        target:
+          - armv7-unknown-linux-gnueabihf
+          - aarch64-unknown-linux-gnu
+          - # Empty target to build for host architecture (x86_64-unknown-linux-gnu) without cross.
```
alsuren, line 19: Can we put the target as x86_64-unknown-linux-gnu here to make it slightly less magical? (even if we *don't* decide use cross for this target)

.github/workflows/package.yml
qwandor: Done. On some recent run it was 184 MB.

run.sh
qwandor: Whoops, done.

.github/workflows/package.yml
qwandor: This doesn't seem to work for me; it runs in the docker container `rustembedded/cross:x86_64-unknown-linux-gnu-0.2.1` and fails.

.github/workflows/package.yml
qwandor: If we can make `cross` work then sure, but if not then I think matching on this target string rather than just an empty string to not use cross is worse.

PR #61, opened 2020-10-26
Add job to create a draft release when a mijia-homie-* tag is pushed.

qwandor:
Debian packages will be built as usual and uploaded to the release as assets. The release can be published manually once it has been checked.

This is the next step for #18.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -3,6 +3,7 @@ name: Package
 on:
   push:
     branches: [master]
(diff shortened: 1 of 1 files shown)


qwandor:
Here's an example of a release created by this job, after publishing: https://github.com/qwandor/mijia-homie/releases/tag/mijia-homie-0.1.0

qwandor:
There is no description added here, but you can add one manually before publishing the release.

PR #62, opened 2020-10-27
Use working-directory rather than cd

qwandor:
I hadn't realised that there is a `working-directory` option for steps. This makes things a bit neater.


alsuren (approved):
yeah, that is a bit cleaner. `working-directory` is infuriating in that it only applies to `run` steps or something, and not custom github action steps that you import from elsewhere.

qwandor:
Yeah, that caught me out too, I thought I would be able to use `actions-rs/cargo@v1` everywhere after all, but no.

PR #63, opened 2020-10-29
Upload Debian packages from releases to Bintray based on filenames.

qwandor:
More work for #18.

I've used 'stable' as the distribution, so you can have a separate distribution for nightly builds if you want.

You'll need to update the Bintray config to enable automatic GPG signing before we use this, as it is required by default by apt.


releasing.md
```diff
@@ -0,0 +1,9 @@
+# Release procedure
```
alsuren, line 1: can we put this as a section in a file with a well-known name, like CONTRIBUTING.md or HACKING.md or something?

alsuren (approved):

releasing.md
qwandor: Moved to docs directory.

PR #64, opened 2020-10-29
Update dependencies

qwandor:
The main one is `rumqttc`, which now has my patch to make `Request` implement `PartialEq`, which simplifies the tests a little.

### DIFF
--- a/homie-controller/src/lib.rs
+++ b/homie-controller/src/lib.rs
@@ -636,7 +636,8 @@ mod tests {
     }
 
     fn expect_subscriptions(requests_rx: &Receiver<Request>, subscription_topics: &[&str]) {
-        let requests: Vec<_> = (0..subscription_topics.len())
+        let requests: Vec<_> = subscription_topics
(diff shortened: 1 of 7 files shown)


homie-controller/src/lib.rs
```diff
@@ -637,18 +637,11 @@ mod tests {
 
     fn expect_subscriptions(requests_rx: &Receiver<Request>, subscription_topics: &[&str]) {
         let requests: Vec<_> = (0..subscription_topics.len())
```
alsuren, line 639: Just spotted this (I might not have re-reviewed properly after you added this helper in the original PR
you're doing `.map(|_| ...)` below, so this line could be written as:

```suggestion
        let requests: Vec<_> = subscription_topics.iter()
```

alsuren (approved):

homie-controller/src/lib.rs
qwandor: Done.

PR #65, opened 2020-10-29
Use all Bluetooth adapters rather than just hci0.

### LINKED ISSUE #58: Use correct hci device in mijia crate
On some devices (e.g. VIM3L Ubuntu), the default Bluetooth adapter is `hci1` rather than `hci0`. However, `mijia` is hardcoded to use `hci0`, so it fails.

The `mijia` crate should automatically detect the correct adapter to use.

qwandor:
This fixes #58 the bug on VIM3L, which for some reason has `hci1` rather than `hci0`.

Is there a better way to do this?

### DIFF
--- a/mijia/src/bluetooth.rs
+++ b/mijia/src/bluetooth.rs
@@ -5,7 +5,7 @@ use core::future::Future;
 use dbus::arg::{RefArg, Variant};
 use dbus::nonblock::stdintf::org_freedesktop_dbus::ObjectManager;
 use dbus::nonblock::SyncConnection;
-use eyre::WrapErr;
+use eyre::{bail, WrapErr};
 use futures::FutureExt;
 use itertools::Itertools;
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


mijia/src/bluetooth.rs
```diff
@@ -109,19 +109,38 @@ impl BluetoothSession {
...
     }
 
-    /// Power on the bluetooth adapter adapter and start scanning for devices.
+    /// Power on all bluetooth adapters and start scanning for devices.
     pub async fn start_discovery(&self) -> Result<(), eyre::Error> {
-        let adapter = dbus::nonblock::Proxy::new(
+        let bluez_root = dbus::nonblock::Proxy::new(
             "org.bluez",
-            "/org/bluez/hci0",
+            "/",
```
alsuren, line 116: We might be able to  start at "/org/bluez" but I haven't checked.

alsuren (approved):

mijia/src/bluetooth.rs
qwandor: That appears not to work.

PR #68, opened 2020-10-31
Docs for mijia crate

qwandor:
Documentation! README!


mijia/src/bluetooth.rs
```diff
@@ -14,7 +14,7 @@ use std::fmt::{self, Display, Formatter};
 use std::str::FromStr;
 use std::sync::Arc;
 
-/// Opaque identifier for a bluetooth device which the system knows about.
+/// Opaque identifier for a Bluetooth device which the system knows about.
```
alsuren, line 17: /// An opaque identifier for the device, including a reference to which adapter it was
    /// discovered on. This can be used to connect to it.

add some of this information here too

alsuren (approved):

mijia/src/bluetooth.rs
qwandor: Done.

PR #72, opened 2020-10-31
Use cross for x86_64 package builds too

qwandor:
You'll need to push the Docker image before this works.

Next step for #18.


alsuren (approved):
assuming it builds on your fork

qwandor:
Yep, it seems to work.

PR #74, opened 2020-10-31
Set firmware name and version properly

qwandor:
This also has my `HomieExtension` trait hacks, which I am less sure about. I can put them in a separate PR if you prefer. The first two commits are a bug fix, the third is my attempt at refactoring to make things more generic.

### DIFF
--- a/homie-device/src/lib.rs
+++ b/homie-device/src/lib.rs
@@ -25,8 +25,6 @@ pub use crate::types::{Datatype, Node, Property};
 
 const HOMIE_VERSION: &str = "4.0";
 const HOMIE_IMPLEMENTATION: &str = "homie-rs";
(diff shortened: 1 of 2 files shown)


homie-device/src/lib.rs
```diff
@@ -516,6 +521,17 @@ impl DevicePublisher {
     }
 }
 
+#[async_trait]
+trait HomieExtension {
+    fn extension_id(&self) -> &'static str;
```
alsuren, line 526: docstring?

alsuren (approved):
I don't think we *need* the trait yet, but it's approximately neutral in terms of complexity and encodes what you mean quite well. It also doesn't affect the public API, so we can iterate on it.

qwandor:
Skipping the refactor for now and just merging the bugfix.

Issue #34, opened 2020-10-17
Homie to InfluxDB exporter


alsuren:
Might be best to set up a local development environment for influxdb+grafana first.

* pick an influxdb library 
  * https://github.com/lautis/ruuvitag-listener just uses the line format

* build it as an example in homie-controller for now

(I also found this in my travels: https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mqtt_consumer )

qwandor-google:
https://crates.io/crates/influx_db_client seems to be the most popular and maintained crate for InfluxDB.

qwandor:
There are some stability issues to fix still, but it is released and working so closing this issue.

PR #75, opened 2020-11-07
Add a Homie to InfluxDB export service

qwandor:
First cut at #34. I found a few other minor issues while I was at it.


qwandor:
For reference, the data in InfluxDB ends up looking like this:

```
> show measurements
name: measurements
name
----
boolean
color
float
integer
> select * from boolean limit 10;
name: boolean
time                device_id     device_name node_id node_name node_type property_id property_name value
----                ---------     ----------- ------- --------- --------- ----------- ------------- -----
1604787659395000000 m5stickc-test M5StickC    light   Light     switch    on          On            false
1604788148155000000 m5stickc-test M5StickC    light   Light     switch    on          On            false
> select * from color limit 10;
name: color
time                device_id     device_name node_id node_name node_type property_id property_name value
----                ---------     ----------- ------- --------- --------- ----------- ------------- -----
1604787659498000000 m5stickc-test M5StickC    light   Light     switch    color       Colour        255,255,255
1604788148168000000 m5stickc-test M5StickC    light   Light     switch    color       Colour        255,255,255
> select * from integer limit 10;
name: integer
time                device_id          device_name               node_id      node_name           node_type    property_id property_name unit value
----                ---------          -----------               -------      ---------           ---------    ----------- ------------- ---- -----
1604787631595000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C138487922 Bedside table       Mijia sensor humidity    Humidity      %    55
1604787632472000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C138487922 Bedside table       Mijia sensor battery     Battery level %    40
1604787633507000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C13821459B Fridge door         Mijia sensor humidity    Humidity      %    44
1604787633732000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C13821459B Fridge door         Mijia sensor battery     Battery level %    69
1604787633850000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C1381E0AE8 Pallet bookshelf    Mijia sensor humidity    Humidity      %    62
1604787633878000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C1381E0AE8 Pallet bookshelf    Mijia sensor battery     Battery level %    67
1604787633936000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C1381F003F Living room shelves Mijia sensor humidity    Humidity      %    60
1604787633953000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C1381F003F Living room shelves Mijia sensor battery     Battery level %    82
1604787634025000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C138E98330 Kitchen shelf       Mijia sensor humidity    Humidity      %    59
1604787634037000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C138E98330 Kitchen shelf       Mijia sensor battery     Battery level %    72
> select * from float limit 10;
name: float
time                device_id          device_name               node_id      node_name           node_type    property_id property_name unit value
----                ---------          -----------               -------      ---------           ---------    ----------- ------------- ---- -----
1604787629469000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C138487922 Bedside table       Mijia sensor temperature Temperature   ºC   22.7
1604787633287000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C13821459B Fridge door         Mijia sensor temperature Temperature   ºC   11.93
1604787633817000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C1381E0AE8 Pallet bookshelf    Mijia sensor temperature Temperature   ºC   21.36
1604787633921000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C1381F003F Living room shelves Mijia sensor temperature Temperature   ºC   22.16
1604787634011000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C138E98330 Kitchen shelf       Mijia sensor temperature Temperature   ºC   22.3
1604787634048000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C138D6D08B Fridge drawer       Mijia sensor temperature Temperature   ºC   0.83
1604787634077000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C138C950F2 Outside chair       Mijia sensor temperature Temperature   ºC   9.89
1604787634106000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C138598607 Anusha desk         Mijia sensor temperature Temperature   ºC   21.91
1604787634137000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C13885646D Table dangly        Mijia sensor temperature Temperature   ºC   10.12
1604787634192000000 mijia-bridge-raspi Raspberry Pi Mijia bridge A4C13859D9A2 Office              Mijia sensor temperature Temperature   ºC   20.9

```

qwandor:
I should also note that the reason I went for a different measurement per datatype is because InfluxDB doesn't allow a field on a measurement to contain values of different types. The other option would be to have one measurement but a different field name depending on the type. I'm not entirely sure which is better.

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+    if let Some(device) = controller.devices().get(&device_id) {
+        if let Some(node) = device.nodes.get(&node_id) {
+            if let Some(property) = node.properties.get(&property_id) {
+                if let Some(point) =
+                    point_for_property_value(device, node, property, SystemTime::now())
+                {
+                    // TODO: What should rp be?
+                    // TODO: Handle errors
+                    influx_db_client
+                        .write_point(point, INFLUXDB_PRECISION, None)
```
alsuren, line 205: Seems like it's possible for the influxdb admin to set a default retention policy for a db, so leaving this as None sounds like the best thing to do. https://docs.influxdata.com/influxdb/v1.8/query_language/manage-database/#create-a-default-retention-policy

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+        point = point.add_tag("node_type", Value::String(node_type))
+    }
+
+    Some(point)
+}
+
+fn simplify_unit_vec<E>(m: Result<Vec<()>, E>) -> Result<(), E> {
+    m.map(|_| ())
+}
+
```
alsuren, line 270: I know that the other binary's main.rs is just as big, but it's shocking seeing all of this code appear in one file in one PR.

Can we split out a `config.rs` and an `influx.rs` so that main.rs only contains `main()` and maybe `spawn_homie_poll_loop()` and `simplify_unit_vec()`?

I can do the mijia-homie refactor as a separate PR if you want.

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+        )
+        .add_field("value", value)
+        .add_tag("device_id", Value::String(device.id.to_owned()))
+        .add_tag("node_id", Value::String(node.id.to_owned()))
+        .add_tag("property_id", Value::String(property.id.to_owned()));
+    if let Some(device_name) = device.name.to_owned() {
+        point = point.add_field("device_name", Value::String(device_name));
+    }
+    if let Some(node_name) = node.name.to_owned() {
+        point = point.add_field("node_name", Value::String(node_name));
```
alsuren, line 252: I think that these should be tags as well: If I want to plot a graph of all fridge temperature measurements, I want node_name and property_name to be indexed (I still haven't worked out how the indexes work in-memory, or how indexes for different tags are combined, so my instincts might be wrong).

We're not expecting either of these things to be high cardinality, so I think tag is probably more appropriate: https://docs.influxdata.com/influxdb/v1.8/concepts/schema_and_data_layout/

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+fn point_for_property_value(
+    device: &Device,
+    node: &Node,
+    property: &Property,
+    timestamp: SystemTime,
+) -> Option<Point> {
+    let datatype = property.datatype?;
+    let value = influx_value_for_homie_property(property)?;
+
+    let mut point = Point::new(&datatype.to_string())
```
alsuren, line 237: So this will be one of Integer, Float, Boolean, String, Enum, Color? Seems suspicious but it might just work ;-).

alsuren (commented):
Sorry for taking so long to get around to this. Couple of suggested refactors. I'll try it out locally now and see how it behaves.

homie-influx/src/main.rs
qwandor: Lowercase, but yes. See my comment on the PR for an example of how the data looks like in InfluxDB.

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    simplify_unit_vec(try_join_all(join_handles).await)
+}
+
+/// Read mappings of the form "homie_prefix:influxdb_database" from the given file, ignoring any
+/// lines starting with '#'.
+fn mappings_from_file(filename: &str) -> Result<Vec<Mapping>, eyre::Error> {
+    let mut mappings = Vec::new();
+    let file = File::open(filename)?;
```
alsuren, line 75:
```suggestion
    let file = File::open(filename).wrap_err_with(|| format!("Failed to open {}", filename))?;

```

homie-influx/src/main.rs
alsuren: When I say "high cardinality", even a low-spec influxdb box will be fine with a series cardinality of 100,000 according to https://docs.influxdata.com/influxdb/v1.8/guides/hardware_sizing/.

See also https://docs.influxdata.com/influxdb/v1.8/concepts/glossary/#series-cardinality - seems like serieses are grouped by `(database, measurement, tag set, field key)` when determining cardinality. We can probably infer something about how the indexes are structured from this. I would assume that each series is recorded as a column, and that you calculate the tag sets that you're looking for first, and then combine the every series that matches your filters into a single series before rendering?

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+}
+
+fn spawn_homie_poll_loop(
+    mut event_loop: HomieEventLoop,
+    controller: Arc<HomieController>,
+    influx_db_client: Client,
+) -> JoinHandle<Result<(), PollError>> {
+    task::spawn(async move {
+        loop {
+            if let Some(event) = controller.poll(&mut event_loop).await? {
```
alsuren, line 152: I'm currently getting:
```
Error: Mqtt state: Last pingreq isn't acked

Caused by:
   0: Mqtt state: Last pingreq isn't acked
   1: Last pingreq isn't acked
```
which I'm assuming is happening here.

Do we want to do something like this to allow us to debug what's going on? 
```
diff --git a/homie-influx/src/main.rs b/homie-influx/src/main.rs
index f451b2e..91a701b 100644
--- a/homie-influx/src/main.rs
+++ b/homie-influx/src/main.rs
@@ -1,8 +1,6 @@
 use futures::future::try_join_all;
 use futures::FutureExt;
-use homie_controller::{
-    Datatype, Device, Event, HomieController, HomieEventLoop, Node, PollError, Property,
-};
+use homie_controller::{Datatype, Device, Event, HomieController, HomieEventLoop, Node, Property};
 use influx_db_client::reqwest::Url;
 use influx_db_client::{Client, Point, Precision, Value};
 use rumqttc::MqttOptions;
@@ -146,10 +144,14 @@ fn spawn_homie_poll_loop(
     mut event_loop: HomieEventLoop,
     controller: Arc<HomieController>,
     influx_db_client: Client,
-) -> JoinHandle<Result<(), PollError>> {
+) -> JoinHandle<Result<(), eyre::Report>> {
     task::spawn(async move {
         loop {
-            if let Some(event) = controller.poll(&mut event_loop).await? {
+            if let Some(event) = controller
+                .poll(&mut event_loop)
+                .await
+                .wrap_err_with(|| format!("Failed to poll {:?}", controller))?
+            {
                 match event {
                     Event::PropertyValueChanged {
                         device_id,
```

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+            datatype: Some(Datatype::Color),
+            settable: false,
+            retained: true,
+            unit: None,
+            format: None,
+            value: Some("12,34,56".to_owned()),
+        };
+        assert_debug_eq(
+            influx_value_for_homie_property(&property).unwrap(),
+            Value::String("12,34,56".to_owned()),
```
alsuren, line 385: I was about to suggest splitting value into red green and blue components, but maybe a string is what we really mean. In practice you don't use all 256^3 colours for LED light: you just use one of about 10 specific value combinations from a colour palette.

homie-influx/src/main.rs
alsuren: If I do that then I get:
```
Error: Failed to poll HomieController { mqtt_client: AsyncClient { request_tx: Sender { .. }, cancel_tx: Sender { .. } }, base_topic: "homie", devices: Mutex { data: {"mijia-bridge": Device { id: "mijia-bridge", homie_version: "4.0", name: Some("Mijia bridge"), state: Lost, implementation: Some("homie-rs"), nodes: {}, extensions: [Extension { id: "org.homie.legacy-stats", version: "0.1.1", homie_versions: ["4.x"] }, Extension { id: "org.homie.legacy-firmware", version: "0.1.1", homie_versions: ["4.x"] }], local_ip: Some("100.115.92.198"), mac: Some("02:42:19:F6:14:A1"), firmware_name: Some("homie-device"), firmware_version: Some("0.3.0"), stats_interval: Some(60s), stats_uptime: Some(480s), stats_signal: None, stats_cputemp: None, stats_cpuload: None, stats_battery: None, stats_freeheap: None, stats_supply: None }} } }
```

(I was running with a modified version of my mijia-homie .env, and hadn't realised that you added the MQTT_ prefix to a bunch of things. I think that means it was connecting to test.mosquitto.org. Waiting for timeouts here is quite time-consuming, so I haven't debugged it much further than that)

homie-influx/src/main.rs
alsuren: I suppose the most compelling reason to make it a tag is that if I patch this field to be a tag then I can do group by `node_name` and alias by `[[tag_node_name]]` and get the following graph with names straight away:

Screenshot 2020-11-09 at 15 28 27


 (I'm not sure whether `explore` urls are portable, but this works on my local grafana instance and might be useful to you as well? http://localhost:3000/explore?orgId=1&left=%5B%22now-5m%22,%22now%22,%22InfluxDB%22,%7B%22datasource%22:%22InfluxDB%22,%22policy%22:%22default%22,%22resultFormat%22:%22time_series%22,%22orderByTime%22:%22ASC%22,%22tags%22:%5B%7B%22key%22:%22property_id%22,%22operator%22:%22%3D%22,%22value%22:%22temperature%22%7D%5D,%22groupBy%22:%5B%7B%22type%22:%22time%22,%22params%22:%5B%22$__interval%22%5D%7D,%7B%22type%22:%22tag%22,%22params%22:%5B%22node_name%22%5D%7D%5D,%22select%22:%5B%5B%7B%22type%22:%22field%22,%22params%22:%5B%22value%22%5D%7D,%7B%22type%22:%22mean%22,%22params%22:%5B%5D%7D%5D%5D,%22measurement%22:%22float%22,%22alias%22:%22%5B%5Btag_node_name%5D%5D%22,%22hide%22:false%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D )

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+fn mappings_from_file(filename: &str) -> Result<Vec<Mapping>, eyre::Error> {
+    let mut mappings = Vec::new();
+    let file = File::open(filename)?;
+    for line in BufReader::new(file).lines() {
+        let line = line?;
+        if !line.starts_with('#') {
+            let parts: Vec<&str> = line.splitn(2, ':').collect();
+            if parts.len() != 2 {
+                eyre::bail!("Invalid line '{}'", line);
+            }
```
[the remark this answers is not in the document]
qwandor: Done.

homie-influx/src/main.rs
qwandor: Hopefully that works as expected! Changed the TODO to a comment.

homie-influx/src/main.rs
qwandor: Okay, done.

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+async fn send_property_value(
+    controller: &HomieController,
+    influx_db_client: &Client,
+    device_id: String,
+    node_id: String,
+    property_id: String,
+) {
+    if let Some(device) = controller.devices().get(&device_id) {
+        if let Some(node) = device.nodes.get(&node_id) {
+            if let Some(property) = node.properties.get(&property_id) {
```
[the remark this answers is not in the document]
qwandor: Lifetimes make it hard to put on `HomieController`. I've made a helper method, but I'm not convinced it is actually more readable.

homie-influx/src/main.rs
qwandor: Done, thanks!

homie-influx/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+
+    Ok(mappings)
+}
+
+/// Construct a new InfluxDB `Client` based on configuration options or defaults.
+///
+/// The database name is not set; make sure to set it before using the client.
+fn get_influxdb_client() -> Result<Client, eyre::Report> {
```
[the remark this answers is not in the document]
qwandor: Done.

homie-influx/src/main.rs
qwandor: I don't think dumping the state of the `HomieController` really adds much here, so just added a static string as context.

homie-influx/src/main.rs
qwandor: Done.

homie-influx/src/main.rs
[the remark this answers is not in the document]
alsuren: Ah I see. You're probably right. I hadn't appreciated that we also use device, node and property lower down, so I was thinking you could just clone the property and be done with it or something.

alsuren (approved):

homie-influx/src/main.rs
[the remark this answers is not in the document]
qwandor: Rolled this back to the original code.

PR #77, opened 2020-11-09
restart always, even if something goes wrong at boot

alsuren:
Just noticed that cottagepi wasn't coming up after a power-cycle. Maybe this will help (following advice from https://selivan.github.io/2017/12/30/systemd-serice-always-restart.html).

Just pushing it to the device with run.sh now.

### DIFF
--- a/mijia-homie.service
+++ b/mijia-homie.service
@@ -2,6 +2,9 @@
 Description=mijia-homie - send temperature readings to mqtt
 After=network.target
 
+# https://selivan.github.io/2017/12/30/systemd-serice-always-restart.html
+StartLimitIntervalSec=0
+
 [Service]
 Type=simple
 User=pi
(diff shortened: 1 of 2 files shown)


mijia-homie.service
```diff
@@ -10,6 +13,7 @@ Environment=RUST_BACKTRACE=1
 Environment=RUST_LIB_BACKTRACE=1
 ExecStart=/home/pi/mijia-homie
 Restart=always
+RestartSec=1
```
qwandor, line 16: I think this should be longer (at least 5 seconds, perhaps 10 seconds or more) to avoid getting everything bogged down with continuous restarts when it keeps failing.

qwandor:
Add this to `mijia-homie/debian-scripts/mijia-homie.service` too.

qwandor (approved):

PR #78, opened 2020-11-10
Various fixes for homie-influx

qwandor:
Some minor improvements to error handling, and a bugfix to use a unique client name per MQTT connection. (Whoops, I didn't test this with more than one mapping before.)

Ref #34.


alsuren (approved):
Looks good.

PR #79, opened 2020-11-14
More tests and clean-ups for homie-influx

qwandor:
I've updated to the latest version of `influx_db_client` with the fixes to implement `PartialEq`, which lets me clean up the existing tests and add some new ones. There's also a couple of other minor fixes in here, for logging and reading mappings.

### DIFF
--- a/homie-influx/src/config.rs
+++ b/homie-influx/src/config.rs
@@ -42,7 +42,7 @@ fn mappings_from_file(filename: &str) -> Result<Vec<Mapping>, eyre::Report> {
     let file = File::open(filename).wrap_err_with(|| format!("Failed to open {}", filename))?;
     for line in BufReader::new(file).lines() {
         let line = line?;
-        if !line.starts_with('#') {
+        if line.len() > 0 && !line.starts_with('#') {
             let parts: Vec<&str> = line.split(':').collect();
             if parts.len() != 2 {
                 eyre::bail!("Invalid line '{}'", line);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -596,9 +596,9 @@ dependencies = [
 
 [[package]]
 name = "influx_db_client"
-version = "0.4.4"
+version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


homie-influx/src/config.rs
```diff
@@ -42,7 +42,7 @@ fn mappings_from_file(filename: &str) -> Result<Vec<Mapping>, eyre::Report> {
     let file = File::open(filename).wrap_err_with(|| format!("Failed to open {}", filename))?;
     for line in BufReader::new(file).lines() {
         let line = line?;
-        if !line.starts_with('#') {
+        if line.len() > 0 && !line.starts_with('#') {
```
alsuren, line 45: I feel like `cargo clippy` will have something to say about this.

alsuren (approved):

homie-influx/src/config.rs
qwandor: It doesn't seem to.

alsuren:
```
    Checking homie-influx v0.1.0 (/Users/alsuren/src/mijia-homie/homie-influx)
warning: length comparison to zero
  --> homie-influx/src/config.rs:29:8
   |
29 |     if mappings.len() == 0 {
   |        ^^^^^^^^^^^^^^^^^^^ help: using `is_empty` is clearer and more explicit: `mappings.is_empty()`
   |
   = note: `#[warn(clippy::len_zero)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_zero

warning: length comparison to zero
  --> homie-influx/src/config.rs:45:12
   |
45 |         if line.len() > 0 && !line.starts_with('#') {
   |            ^^^^^^^^^^^^^^ help: using `!is_empty` is clearer and more explicit: `!line.is_empty()`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_zero

warning: 2 warnings emitted

    Finished dev [unoptimized + debuginfo] target(s) in 51.29s
```

note that clippy shares its message cache with `cargo check` (despite https://github.com/rust-lang/cargo/issues/8075 being closed) so you need to force a rebuild to make it tell you anything (either with a `cargo clean --package homie-influx` beforehand or a noop feature flag, as described in the last comment of that bug).

PR #80, opened 2020-11-15
Add Clippy to CI and fix issues

alsuren:


homie-controller/src/lib.rs
```diff
@@ -124,6 +124,12 @@ impl HomieEventLoop {
     }
 }
 
+struct PublishResponse {
```
qwandor, line 127: At least derive `Debug`, might as well also derive `Eq`, `PartialEq`, `Clone`.

alsuren: I doubt that we will ever need to debug this, since it's an internal struct that lives for about 2 lines before being consumed.

I will add a docstring saying what it's used for, and we can add derives when we need them for tests or make the function public.

qwandor: It might come in handy for tests in the future, but I guess I can always add it then.

qwandor (approved):

qwandor:
Please rebase to apply your fixup before merging.

PR #81, opened 2020-11-17
homie-controller: Update version number for release.

qwandor:
The version of rumqttc was updated, which breaks compatibility of the public API.


alsuren (approved):
Seems sensible. I wonder if there is any tooling to automatically catch things like this.

PR #86, opened 2020-11-29
Add methods to get historical readings

### LINKED ISSUE #49: Add methods to Mijia library for reading and writing various device settings
Apart from just subscribing to temperature and humidity readings, there are GATT characteristics exposed for various settings. It would be good to support them in the `mijia` crate.

qwandor:
Fixes #49.

### DIFF
--- a/mijia/src/decode/history.rs
+++ b/mijia/src/decode/history.rs
@@ -1,6 +1,9 @@
+use crate::decode::time::decode_time;
 use crate::decode::{check_length, DecodeError};
 use std::convert::TryInto;
+use std::fmt::{self, Display, Formatter};
 use std::ops::Range;
+use std::time::SystemTime;
 
 /// Decode a range of indices encoded as a last index and count into a Rust half-open `Range`.
 pub(crate) fn decode_range(value: &[u8]) -> Result<Range<u32>, DecodeError> {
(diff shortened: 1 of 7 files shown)


mijia/src/decode/history.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    let end = last_index + 1;
+    let start = end - count;
+
+    Ok(start..end)
+}
+
+/// A historical temperature/humidity record stored by a sensor.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Record {
```
alsuren, line 23: HistoricalRecord maybe?

mijia/src/lib.rs
```diff
@@ -25,9 +28,16 @@ const CONNECTION_INTERVAL_CHARACTERISTIC_PATH: &str = "/service0021/char0045";
 const CLOCK_CHARACTERISTIC_PATH: &str = "/service0021/char0022";
 const TEMPERATURE_UNIT_CHARACTERISTIC_PATH: &str = "/service0021/char0032";
 const COMFORT_LEVEL_CHARACTERISTIC_PATH: &str = "/service0021/char0042";
+const HISTORY_RANGE_CHARACTERISTIC_PATH: &str = "/service0021/char0025";
+const HISTORY_DELETE_CHARACTERISTIC_PATH: &str = "/service0021/char003f";
+const HISTORY_LAST_RECORD_CHARACTERISTIC_PATH: &str = "/service0021/char002b";
+const HISTORY_INDEX_CHARACTERISTIC_PATH: &str = "/service0021/char0028";
+const HISTORY_RECORDS_CHARACTERISTIC_PATH: &str = "/service0021/char002e";
```
alsuren, line 35: what does this look like if it is sorted by path?

mijia/src/lib.rs
```diff
@@ -192,36 +231,133 @@ impl MijiaSession {
...
+
+    /// Try to get all historical records for the sensor.
+    pub async fn get_all_history(&self, id: &DeviceId) -> Result<Vec<Option<Record>>, MijiaError> {
+        let history_range = self.get_history_range(&id).await?;
+        // TODO: Get event stream that is filtered by D-Bus.
+        let (msg_match, events) = self.event_stream().await?;
+        let mut events = events.timeout(HISTORY_RECORD_TIMEOUT);
+        self.start_notify_history(&id, Some(0)).await?;
+
+        let mut history = vec![None; history_range.len()];
```
alsuren, line 303: is history_range start always 0?

mijia/src/lib.rs
```diff
@@ -192,36 +231,133 @@ impl MijiaSession {
...
+                        } else {
+                            log::error!(
+                                "Got record {:?} for sensor {:?} out of bounds {:?}",
+                                record,
+                                id,
+                                history_range
+                            );
+                        }
+                    } else {
+                        log::warn!("Got record for wrong sensor {:?}", record_id);
```
alsuren, line 324: I expect these to be quite noisy in some cases. Probably fine for now though.

alsuren (approved):

mijia/src/lib.rs
qwandor: As good as any other way. Done.

mijia/src/decode/history.rs
qwandor: Renamed to `HistoryRecord`, to match the enum variant.

mijia/src/lib.rs
qwandor: That should only happen if we try to retrieve history for two sensors simultaneously, and it will be fixed once we do the TODO above to filter the D-Bus stream.

PR #88, opened 2020-12-01
Handle Homie property datatypes better in homie-device

### LINKED ISSUE #35: Improve handling of Homie typed properties

qwandor:
This adds separate datatype-specific constructors to `Property` which take the appropriate type of format (if any) rather than an arbitrary string, and types for parsing colour values.

Fixes #35.

### DIFF
--- a/homie-device/src/types.rs
+++ b/homie-device/src/types.rs
@@ -1,4 +1,5 @@
-use std::fmt::Debug;
+use std::fmt::{self, Debug, Display, Formatter};
+use std::ops::Range;
 
 /// The data type for a Homie property.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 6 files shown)


homie-device/src/types.rs
```diff
@@ -58,33 +100,181 @@ impl Property {
         datatype: Datatype,
         settable: bool,
         unit: Option<&str>,
-        format: Option<&str>,
+        format: Option<String>,
```
alsuren, line 103: This is a bit inconsistent. Is the intention to make this function private?

qwandor: Fair point. I was thinking to keep this function public, in case anyone wants a generic way to create properties. So there are two options that seem plausible to me:
1. Add a private `Property::make` method or something like that which does what this function does, and have everything else call that. Change the signature of `Property::new` back to what it used to be.
2. Change `Property::new` to use `impl Into<String>` for all relevant parameters. The problem with this is that passing `None` as an `Option< impl Into<String>>` can't have its type inferred, so callers have to explicitly choose something (and it doesn't matter what).

Any idea which is more idiomatic?

alsuren: Probably 1. When in doubt, public functions should accept &str and do the clone themselves. Maybe call it from_fields() or something?

qwandor: Done.

homie-device/src/types.rs
```diff
@@ -59,32 +101,198 @@ impl Property {
...
+    /// * `unit`: The unit for the property, if any. This may be one of the
+    ///   [recommended units](https://homieiot.github.io/specification/#property-attributes), or
+    ///   any other custom unit.
+    /// * `format`: The possible values for the enum.
+    pub fn enumeration(
+        id: &str,
+        name: &str,
+        settable: bool,
+        unit: Option<&str>,
+        format: &[&str],
```
qwandor, line 213: Is this the right type to use? I tried to do something like `impl IntoIterator<&'a str>` as recommended by https://rust-lang.github.io/api-guidelines/flexibility.html#c-generic but I couldn't get it to work with the example passing in a slice.

alsuren: Sounds okay according to the seed style guide, I think. I suppose if someone has a Vec<String> then they will need to do some conversion, so **maybe** intoiter would be better, but I'm not always convinced by super-generic interfaces.

> Use references where possible - e.g. rather use &str instead of String; or &[T] instead of Vec.
-- https://seed-rs.org/0.8.0/rust

Let me take a quick look at the example but I think this seems fine.

alsuren: Yeah, this feels right. You should always be passing in a slice of static strings to this method, so this will always be the most convenient type.

Ship it.

qwandor: Done.

PR #89, opened 2020-12-02
Add example to subscribe to sensor readings.

qwandor:
Part of #21.

### DIFF
--- a/mijia/examples/readings.rs
+++ b/mijia/examples/readings.rs
@@ -13,23 +13,7 @@ const SCAN_DURATION: Duration = Duration::from_secs(5);
 async fn main() -> Result<(), Report> {
     pretty_env_logger::init();
 
-    // If at least one command-line argument is given, we will only try to connect to sensors whose
-    // MAC address containts one of them as a sub-string.
-    let mut args = std::env::args();
-    let binary_name = args
-        .next()
-        .ok_or_else(|| eyre::eyre!("Binary name missing"))?;
-    let filters: Vec<_> = args.collect();
-
-    if filters
-        .iter()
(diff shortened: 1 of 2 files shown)


mijia/examples/readings.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    if filters
+        .iter()
+        .any(|f| f.contains(|c: char| !(c.is_ascii_hexdigit() || c == ':')))
+    {
+        eprintln!("Invalid MAC addresses {:?}", filters);
+        eprintln!("Usage:");
+        eprintln!("  {} [MAC address]...", binary_name);
+        exit(1);
+    }
```
alsuren, line 32: Maybe split this out into a function since it's not that interesting to the example (I feel like there is a copy-pasted version of this in another example. That could also be split out into a function, but it's fine for them to both be copy-paste)

mijia/examples/readings.rs
```diff
@@ -0,0 +1,78 @@
...
+                println!("{:?}: {}", id, readings);
+            }
+            _ => println!("Event: {:?}", event),
+        }
+    }
+    session
+        .bt_session
+        .connection
+        .remove_match(msg_match.token())
+        .await?;
```
alsuren, line 70: ugh. I *really* want to make a newtype that wraps `events` and `msg_match` and make it do `tokio::spawn(|| remove_match())` as a drop-guard or something.

alsuren (approved):
Looks like a good example. Split out the function and then ship it.

mijia/examples/readings.rs
qwandor: Done.

PR #90, opened 2020-12-05
Support multiple Bluetooth adapters

### LINKED ISSUE #84: Support using multiple Bluetooth adapters simultaneously
The `mijia` crate should already discover sensors on multiple adapters, we just need to add some state logic to `mijia-homie` to avoid trying to connect to the same sensor via two different adapters at once. The first to successfully connect should win, I guess.

qwandor:
Try to connect to sensors on all adapters they are discovered on, until one manages to connect.

Fixes #84.

### DIFF
--- a/mijia-homie/src/main.rs
+++ b/mijia-homie/src/main.rs
@@ -1,7 +1,9 @@
 #![type_length_limit = "1138969"]
 
 use backoff::{future::FutureOperation, ExponentialBackoff};
-use futures::stream::StreamExt;
+use eyre::{eyre, Report};
+use futures::stream::{self, StreamExt};
+use futures::Stream;
 use futures::TryFutureExt;
 use homie_device::{HomieDevice, Node, Property};
 use itertools::Itertools;
@@ -103,7 +105,7 @@ fn get_mqtt_options(device_id: &str) -> MqttOptions {
     mqtt_options
 }
 
-#[derive(Debug, Copy, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
 enum ConnectionStatus {
     /// Not yet attempted to connect. Might already be connected from a previous
     /// run of this program.
@@ -119,18 +121,16 @@ enum ConnectionStatus {
     /// events might be received racily. The sensor might actually be Connected.
     MarkedDisconnected,
     /// Connected and subscribed to updates
-    Connected,
-    /// Connected on a different Bluetooth adapter, on a system which has more than one.
-    ConnectedOnOtherAdapter,
+    Connected { id: DeviceId },
 }
 
 #[derive(Debug, Clone)]
 struct Sensor {
-    id: DeviceId,
     mac_address: MacAddress,
     name: String,
     last_update_timestamp: Instant,
     connection_status: ConnectionStatus,
+    ids: Vec<DeviceId>,
 }
 
 impl Sensor {
@@ -144,11 +144,11 @@ impl Sensor {
             .cloned()
             .unwrap_or_else(|| props.mac_address.to_string());
         Self {
-            id: props.id,
             mac_address: props.mac_address,
             name,
             last_update_timestamp: Instant::now(),
             connection_status: ConnectionStatus::Unknown,
+            ids: vec![props.id],
         }
     }
 
@@ -216,14 +216,15 @@ impl Sensor {
         Ok(())
     }
 
-    async fn mark_connected(&mut self, homie: &mut HomieDevice) -> Result<(), eyre::Report> {
+    async fn mark_connected(
+        &mut self,
+        homie: &mut HomieDevice,
+        id: DeviceId,
+    ) -> Result<(), eyre::Report> {
+        assert!(self.ids.contains(&id));
         homie.add_node(self.as_node()).await?;
-        self.connection_status = ConnectionStatus::Connected;
+        self.connection_status = ConnectionStatus::Connected { id };
         Ok(())
-    }
-
-    fn name_with_adapter(&self) -> String {
-        format!("{}/{}", self.name, self.id.adapter())
     }
 }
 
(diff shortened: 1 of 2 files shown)


mijia-homie/src/main.rs
```diff
@@ -120,6 +120,8 @@ enum ConnectionStatus {
     MarkedDisconnected,
     /// Connected and subscribed to updates
     Connected,
+    /// Connected on a different Bluetooth adapter, on a system which has more than one.
+    ConnectedOnOtherAdapter,
 }
 
 #[derive(Debug, Clone)]
```
alsuren, line 127: I can believe that your approach will *work*, but I think it would be a better model of reality to make `Sensor::id` plural somehow. I realise that this might be a bunch more work though, so I'm happy to merge this as an intermediate step. Probably best to sleep on it.

qwandor: Pushed a first cut at this, with some TODOs still. We might want to change the interface of the `mijia` crate too.

mijia-homie/src/main.rs
```diff
@@ -289,19 +295,20 @@ async fn bluetooth_connection_loop(
 
         // Check the state of each sensor and act on it if appropriate.
         {
-            let ids: Vec<DeviceId> = state.lock().await.sensors.keys().cloned().collect();
-            for id in ids {
+            let mac_addresses: Vec<MacAddress> =
+                state.lock().await.sensors.keys().cloned().collect();
```
alsuren, line 299: (maybe later) would be fun to use a MacAddress type that is Copy and have this be `state.lock().await.sensors.keys().copied().collect()`

mijia-homie/src/main.rs
```diff
@@ -289,19 +295,20 @@ async fn bluetooth_connection_loop(
...
                 let connection_status = state
                     .lock()
                     .await
                     .sensors
-                    .get(&id)
+                    .get(&mac_address)
                     .map(|sensor| {
                         log::trace!("State of {} is {:?}", sensor.name, sensor.connection_status);
-                        sensor.connection_status
+                        sensor.connection_status.to_owned()
```
alsuren, line 308: is the adapter id (that's in the DeviceId string) always an integer? If so, maybe we *could* make DeviceId Copy, and convert it to a string only when we need to?

mijia-homie/src/main.rs
```diff
@@ -396,42 +417,64 @@ async fn connect_sensor_with_id(
...
-        .wrap_err_with(|| format!("connecting to {:?}", id))?;
+    name: &str,
+    ids: Vec<DeviceId>,
+) -> Result<DeviceId, eyre::Report> {
+    // Try to connect to the ids in turn, and get the first one that succeeds. If they all fail then
+    // return an error.
+    let id = first_ok(
+        stream::iter(ids)
+            .then(move |id| async move { session.bt_session.connect(&id).await.map(|()| id) }),
+    )
```
alsuren, line 444: just use `for id in ids {... session.bt_session.connect(&id).await ...}` here. No need to mess around with stream combinators anymore, now that we can .await things in a loop.

mijia-homie/src/main.rs
```diff
@@ -484,13 +527,14 @@ async fn handle_bluetooth_event(
     let sensors = &mut state.sensors;
     match event {
         MijiaEvent::Readings { id, readings } => {
-            if let Some(sensor) = sensors.get_mut(&id) {
+            if let Some(sensor) = get_mut_sensor_by_id(sensors, &id) {
                 sensor.publish_readings(homie, &readings).await?;
                 match sensor.connection_status {
-                    ConnectionStatus::Connected | ConnectionStatus::Connecting { .. } => {}
+                    // TODO: What should we do if the id doesn't match?
+                    ConnectionStatus::Connected { .. } | ConnectionStatus::Connecting { .. } => {}
```
alsuren, line 534: If the adaptor id doesn't match then it's still fine I think. In the worst case, that's the last one, and the watchdog timeout will kick in eventually. In the best case, we keep receiving readings from the wrong interface and don't care other than for debugging

mijia-homie/src/main.rs
```diff
@@ -310,14 +317,22 @@ async fn bluetooth_connection_loop(
...
+    sensors: HashMap<MacAddress, Sensor>,
     homie: HomieDevice,
 }
 
+/// Get the sensor entry for the given id, if any.
+fn get_mut_sensor_by_id<'a>(
+    sensors: &'a mut HashMap<MacAddress, Sensor>,
+    id: &DeviceId,
+) -> Option<&'a mut Sensor> {
+    sensors.values_mut().find(|sensor| sensor.ids.contains(id))
```
alsuren, line 329: can we do `sensors.get_mut(id.mac_address())` or something inline and kill off this function?

mijia-homie/src/main.rs
```diff
@@ -396,42 +417,64 @@ async fn connect_sensor_with_id(
     Ok(())
 }
 
-async fn connect_and_subscribe_sensor_or_disconnect<'a>(
+// TODO: Does this already exist somewhere?
+/// Return the first Ok element of the stream, or a vector of all errors if they all fail.
+async fn first_ok<V, E>(stream: impl Stream<Item = Result<V, E>>) -> Result<V, Vec<E>> {
```
alsuren, line 422: kill this and use a for loop below instead.

alsuren (commented):
There are some ergonomic improvements that could come from making DeviceId a bit more clever, but not essential. Clean up the stream crazyness and then ship it?

mijia-homie/src/main.rs
qwandor: Having `MacAddress`, which is essentially a string, be `Copy` feels wrong to me.

mijia-homie/src/main.rs
qwandor: It's a string. In the cases I've seen so far it's something like `hci0`. How would you store `DeviceId` instead?

mijia-homie/src/main.rs
qwandor: I think that makes sense. I'll log it, but otherwise do nothing.

mijia-homie/src/main.rs
qwandor: Done, though I'm sad about it.

mijia-homie/src/main.rs
qwandor: I'm still not really convinced that parsing BlueZ D-Bus paths is the way to go, it feels quite brittle.

mijia-homie/src/main.rs
qwandor: Done.

alsuren (approved):

PR #92, opened 2020-12-11
Use TOML for config files rather than .env

### LINKED ISSUE #33: toml-based configuration file rather than sensor_names.conf + .env (maybe using https://lib.rs/crates/viperus?)

qwandor:
Fixes #33.


homie-influx/mappings.example.toml
```diff
@@ -0,0 +1,3 @@
+[[mapping]]
```
alsuren, line 1: `[[mappings]]` is probably more what you mean. It's just chance that toml's array of tables syntax is a bit mental

If someone spelt it like this then it is clear that plural is correct:
```
mappings = [
    {homie_prefix="homie", influxdb_database="test"},
    {homie_prefix="homie2", influxdb_database="test2"},
]
```

(spoilers, I think toml is a bit mental)

homie-influx/src/config.rs
```diff
@@ -1,105 +1,158 @@
...
 const DEFAULT_INFLUXDB_URL: &str = "http://localhost:8086";
-const DEFAULT_MAPPINGS_FILENAME: &str = "mappings.conf";
+const CONFIG_FILENAME: &str = "homie_influx.toml";
+const DEFAULT_MAPPINGS_FILENAME: &str = "mappings.toml";
+
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default)]
+pub struct Config {
+    pub mqtt: MqttConfig,
+    pub homie: HomieConfig,
```
alsuren, line 25: If the `[homie]` section is missing, it defaults to `mappings_filename="mappings.toml"`?

homie-influx/src/config.rs
```diff
@@ -1,105 +1,158 @@
...
+#[serde(default)]
+pub struct InfluxDBConfig {
+    #[serde(deserialize_with = "de_from_str")]
+    pub url: Url,
+    pub username: Option<String>,
+    pub password: Option<String>,
+}
+
+/// Deserialize a FromStr by deserializing it as a string then parsing it.
+fn de_from_str<'de, D: Deserializer<'de>, T: FromStr>(d: D) -> Result<T, D::Error>
```
alsuren, line 85: This is used exactly once. Why is it so mental?

mijia-homie/src/main.rs
```diff
@@ -235,8 +192,7 @@ async fn run_sensor_system(
     mut homie: HomieDevice,
     session: &MijiaSession,
 ) -> Result<(), eyre::Report> {
-    let sensor_names = hashmap_from_file(SENSOR_NAMES_FILENAME)
-        .wrap_err(format!("reading {}", SENSOR_NAMES_FILENAME))?;
+    let sensor_names = read_sensor_names()?;
```
alsuren, line 195: suddenly feels odd that this isn't passed in from the top

homie-influx/src/config.rs
qwandor: Yes.

homie-influx/src/config.rs
qwandor: I couldn't see any utility that already existed for this, so I wrote my own.

alsuren: (I wonder if the Url type is secretly already deserialisable if you apply the right feature flags or something. Is it the same type as the one mentioned in here https://github.com/servo/rust-url/issues/420#issuecomment-512239238 or it it reqwest specific?)

mijia-homie/src/main.rs
qwandor: Done.

homie-influx/mappings.example.toml
qwandor: Done.

homie-influx/src/config.rs
qwandor: Turns out it does! That's nicer.

mijia-homie/src/config.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct HomieConfig {
+    pub device_id: String,
+    pub device_name: String,
+    pub prefix: String,
+    pub sensor_names_filename: String,
```
alsuren, line 68: add this to mijia-homie/mijia_homie.example.toml ?

alsuren (approved):

mijia-homie/src/config.rs
qwandor: Done.

PR #97, opened 2020-12-14
Add config option to limit frequency of updates sent to server.

### LINKED ISSUE #94: Add config option to limit frequency of sensor updates
To conserve bandwidth (and database space, pre-aggregation) it would be useful to have a config option to limit the frequency of sensor property updates sent to the MQTT server.

qwandor:
Fixes #94.

Also adds some explanatory comments to the example config files, and fixes a minor logging bug.

### DIFF
--- a/mijia-homie/mijia-homie.example.toml
+++ b/mijia-homie/mijia-homie.example.toml
@@ -3,6 +3,11 @@ device_id="mijia-bridge"
 device_name="Mijia bridge"
 prefix="homie"
 sensor_names_filename="sensor-names.toml"
+# The minimum time to wait between sending consecutive readings for the same sensor.
+# The sensors themselves send updates every 6 seconds or so, so setting this to just under a
+# multiple of 6 will give the most consistent results. 0 means that all sensor updates will be sent
+# to the MQTT broker.
+min_update_period_seconds=0
 
 [mqtt]
 host="test.mosquitto.org"
--- a/mijia-homie/mijia-homie.example.toml
+++ b/mijia-homie/mijia-homie.example.toml
@@ -1,7 +1,11 @@
 [homie]
+# The ID to use for the Homie device. This must be unique for a given prefix and server.
 device_id="mijia-bridge"
+# The human-readable name to use for the Homie device.
 device_name="Mijia bridge"
+# The Homie base MQTT topic.
 prefix="homie"
+# The name of the file containing sensor MAC address to name mappings.
 sensor_names_filename="sensor-names.toml"
 # The minimum time to wait between sending consecutive readings for the same sensor.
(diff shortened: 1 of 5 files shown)


mijia-homie/src/main.rs
```diff
@@ -153,28 +162,39 @@ impl Sensor {
         &mut self,
```
alsuren, line 162: do we have tests for this function?

mijia-homie/src/config.rs
```diff
@@ -66,6 +68,17 @@ pub struct HomieConfig {
     pub device_name: String,
     pub prefix: String,
     pub sensor_names_filename: String,
+    /// The minimum time to wait between sending consecutive readings for the same sensor.
+    #[serde(
+        deserialize_with = "de_duration_seconds",
+        rename = "min_update_period_seconds"
+    )]
+    pub min_update_period: Duration,
```
alsuren, line 76: I was just thinking about interval vs period as names for this thing. I usually think of period for things that are regular, but if the period of updates is 6s and you set this to 10s then your updates will happen at:
0s
not 6s
12s
not 18s
... I guess periodic is correct. Ship it.

mijia-homie/mijia-homie.example.toml
```diff
@@ -1,13 +1,26 @@
...
+# The ID to use for the Homie device. This must be unique for a given prefix and server.
 device_id="mijia-bridge"
+# The human-readable name to use for the Homie device.
 device_name="Mijia bridge"
+# The Homie base MQTT topic.
 prefix="homie"
+# The name of the file containing sensor MAC address to name mappings.
 sensor_names_filename="sensor-names.toml"
+# The minimum time to wait between sending consecutive readings for the same sensor.
+min_update_period_seconds=0
```
alsuren, line 11: Worth noting here that the sensor is known to give updates every 6 seconds, so in practice you should pick a number slightly below the multiple-of-6 that you want, for consistent results.

mijia-homie/src/main.rs
```diff
@@ -460,21 +483,26 @@ async fn service_bluetooth_event_queue(
 async fn handle_bluetooth_event(
     state: Arc<Mutex<SensorState>>,
     event: MijiaEvent,
+    min_update_period: Duration,
```
alsuren, line 486: (optional) I'd be tempted to dump min_update_period on SensorState or something

alsuren (approved):
Looks good. Minor comments, all completely optional.

mijia-homie/src/main.rs
qwandor: Done, though I'm not sure if it's better or worse.

PR #98, opened 2020-12-16
Reconnect to MQTT broker if connection is dropped

qwandor:
This should fix the `homie-influx` part of #7. It also makes errors writing to InfluxDB non-fatal. There is no point crashing just because one database isn't working, as the others may still work.

### DIFF
--- a/homie-influx/src/config.rs
+++ b/homie-influx/src/config.rs
@@ -3,15 +3,18 @@ use influx_db_client::reqwest::Url;
 use influx_db_client::Client;
 use rumqttc::MqttOptions;
 use rustls::ClientConfig;
+use serde::{Deserialize as _, Deserializer};
 use serde_derive::Deserialize;
 use stable_eyre::eyre;
 use stable_eyre::eyre::WrapErr;
 use std::fs::read_to_string;
 use std::sync::Arc;
+use std::time::Duration;
 
 const DEFAULT_MQTT_CLIENT_PREFIX: &str = "homie-influx";
 const DEFAULT_MQTT_HOST: &str = "test.mosquitto.org";
 const DEFAULT_MQTT_PORT: u16 = 1883;
+const DEFAULT_MQTT_RECONNECT_INTERVAL: Duration = Duration::from_secs(5);
 const DEFAULT_INFLUXDB_URL: &str = "http://localhost:8086";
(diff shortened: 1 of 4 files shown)


homie-influx/src/main.rs
```diff
@@ -31,64 +32,86 @@ async fn main() -> Result<(), eyre::Report> {
...
+            influxdb_client,
+            config.mqtt.reconnect_interval,
+        );
         controller.start().await?;
-        join_handles.push(handle.map(|res| Ok(res??)));
+        join_handles.push(handle);
     }
 
-    simplify_unit_vec(try_join_all(join_handles).await)
+    simplify_unit_vec(try_join_all(join_handles).await)?;
```
alsuren, line 45: No need to simplify it you are throwing the result away

homie-influx/src/config.rs
```diff
@@ -3,15 +3,18 @@ use influx_db_client::reqwest::Url;
...
 use stable_eyre::eyre;
 use stable_eyre::eyre::WrapErr;
 use std::fs::read_to_string;
 use std::sync::Arc;
+use std::time::Duration;
 
 const DEFAULT_MQTT_CLIENT_PREFIX: &str = "homie-influx";
 const DEFAULT_MQTT_HOST: &str = "test.mosquitto.org";
 const DEFAULT_MQTT_PORT: u16 = 1883;
+const DEFAULT_MQTT_RECONNECT_INTERVAL: Duration = Duration::from_secs(5);
```
alsuren, line 17: How long does it take to get synced up with all the mqtt state? If you were in a fresh loop, what fraction of your time would be spent hammering the server?

alsuren (approved):
Looks sensible. Hopefully it fixes your dad's problem.

homie-influx/src/config.rs
qwandor: The state is preserved on the broker when `clean_session` is set to false, so the client just needs to authenticate and connect/conack. WIthout a delay it just seems to constantly try to reconnect. I didn't check CPU usage, but there was certainly a lot of log spam.

homie-influx/src/main.rs
qwandor: Done.

PR #106, opened 2020-12-28
Look up characteristics by UUID rather than relying on paths being consistent.

qwandor:
This also moves the first part of `event_stream()` into `BluetoothSession` and adds a few small tests.


mijia/src/bluetooth.rs
```diff
@@ -64,6 +81,53 @@ impl DeviceId {
...
+            object_path: object_path.to_owned(),
+        }
+    }
+
+    /// Get the ID of the device on which this service was advertised.
+    pub fn device(&self) -> DeviceId {
+        let index = self
+            .object_path
+            .rfind('/')
+            .expect("ServiceId object_path doesn't contain a slash.");
```
alsuren, line 102:
```suggestion
            .expect("ServiceId object_path must contain a slash.");
```
and also below.

I find this type of `expect()` message really hard to read, because my brain always reads it as "expect that ServiceId object_path doesn't contain a slash." which is the opposite of what we are actually expecting. 

The code is much easier to read if you state your expectation/invariant in the message as a MUST statement, and the error message is still easy to read if the thing explodes.

("expect" is a very unfortunate name for this function, but this is the world that we live in)

mijia/src/bluetooth.rs
```diff
@@ -229,7 +312,153 @@ impl BluetoothSession {
...
-    fn device(&self, id: &DeviceId) -> impl OrgBluezDevice1 {
+    /// Get a list of all GATT services which the given Bluetooth device offers.
+    ///
+    /// Note that this won't be filled in until the device is connected.
+    pub async fn get_services(
+        &self,
+        device: &DeviceId,
+    ) -> Result<Vec<ServiceInfo>, BluetoothError> {
+        let introspection_xml = self.device(device).introspect().compat().await?;
+        let device_node: Node = serde_xml_rs::from_str(&introspection_xml)?;
```
alsuren, line 323: (optional) DBusNode or something?

mijia/src/bluetooth.rs
```diff
@@ -114,6 +178,25 @@ pub struct DeviceInfo {
...
+    pub primary: bool,
+}
+
+/// Information about a GATT characteristic on a Bluetooth device.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct CharacteristicInfo {
+    /// An opaque identifier for the characteristic on the device, including a reference to which
+    /// adapter it was discovered on.
+    pub id: CharacteristicId,
+    pub uuid: String,
```
alsuren, line 197: there is a uuid crate. Could we type this as
```
struct CharacteristicUuid(uuid::Uuid)
```
?

mijia/src/bluetooth.rs
```diff
@@ -229,7 +312,153 @@ impl BluetoothSession {
...
+
+    /// Convenience method to get a GATT charactacteristic with the given UUID advertised by a
+    /// device as part of the given service.
+    ///
+    /// This is equivalent to calling `get_service_by_uuid` and then `get_characteristic_by_uuid`.
+    pub async fn get_service_characteristic_by_uuid(
+        &self,
+        device: &DeviceId,
+        service_uuid: &str,
+        characteristic_uuid: &str,
```
alsuren, line 409: *_uuid: &str makes me wary. We had a bunch of `uuid: uuid::Uuid` in the futurenhs project, and it caused us problems with positional arguments in the wrong order. A newtype ServiceUuid(Uuid) would avoid this class of problem.

mijia/src/bluetooth_event.rs
```diff
@@ -29,30 +29,32 @@
 use dbus::{arg::cast, arg::RefArg, arg::TypeMismatchError, arg::Variant, Message};
 use std::collections::HashMap;
 
+use crate::bluetooth::{CharacteristicId, DeviceId};
+
 #[derive(Clone, Debug)]
 pub enum BluetoothEvent {
     Powered {
-        object_path: String,
+        adapter: String,
```
alsuren, line 37: AdapterID?

mijia/src/introspect.rs
```diff
@@ -0,0 +1,215 @@
...
+        name="org.freedesktop.DBus.ObjectManager"><method 
+        name="GetManagedObjects"><arg name="objects" type="a{oa{sa{sv}}}" 
+        direction="out"/>
+        </method><signal name="InterfacesAdded"><arg name="object" type="o"/>
+        <arg name="interfaces" type="a{sa{sv}}"/>
+        </signal>
+        <signal name="InterfacesRemoved"><arg name="object" type="o"/>
+        <arg name="interfaces" type="as"/>
+        </signal>
+        </interface><node name="org"/></node>"#,
```
alsuren, line 129: I know this is straight off the wire, but maybe dump this through `xmllint --format -` like we do in `bluez-generated/introspect.sh` (you could even `include_str!()` something in bluez-generated/specs/ if you were feeling fancy)

alsuren (approved):
This is definitely an improvement. Might be worth pinning down the types of the UUIDs and adapter id a bit more, but that can happen in another PR.

mijia/src/bluetooth.rs
qwandor: They are in the `introspect` module, which I think is enough of a namespace.

mijia/src/introspect.rs
qwandor: Done.

mijia/src/bluetooth_event.rs
qwandor: Done.

mijia/src/bluetooth.rs
qwandor: Done.

mijia/src/bluetooth.rs
qwandor: I'm not really keen on making a new type, as it really is just a UUID, but I've used the `uuid` crate.

PR #107, opened 2020-12-28
Move first part of event_stream to BluetoothSession

qwandor:
This also has a few other minor fixes and an extra test.

### DIFF
--- a/mijia-homie/src/main.rs
+++ b/mijia-homie/src/main.rs
@@ -459,19 +459,15 @@ async fn service_bluetooth_event_queue(
     session: &MijiaSession,
 ) -> Result<(), eyre::Report> {
     println!("Subscribing to events");
-    let (msg_match, mut events) = session.event_stream().await?;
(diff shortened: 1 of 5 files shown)


mijia-homie/src/main.rs
```diff
@@ -459,18 +459,13 @@ async fn service_bluetooth_event_queue(
...
         handle_bluetooth_event(state.clone(), event).await?
     }
 
-    session
-        .bt_session
-        .connection
-        .remove_match(msg_match.token())
-        .await?;
     // This should be unreachable, because the events Stream should never end,
     // unless something has gone horribly wrong (or msg_match got dropped?)
```
alsuren, line 470:
```suggestion
    // unless something has gone horribly wrong.
```

alsuren (approved):
💌

mijia-homie/src/main.rs
qwandor: Done.

Issue #91, opened 2020-12-09
bluez-generated: Error: DbusError(D-Bus error: Method "WriteValue" with signature "aya{sv}" on interface "org.bluez.GattCharacteristic1" doesn't exist


ruediger:
Hello,

I'm trying to use bluez-generated to talk to a meat thermometer. Thank you for writing buez-generated and I also borrowed your bluetooth.rs code. However I seem to have trouble writing to a GattCharacteristic. After finding and connecting to the device, I'm trying to call WriteValue. But I get an error

```
Error: DbusError(D-Bus error: Method "WriteValue" with signature "aya{sv}" on interface "org.bluez.GattCharacteristic1" doesn't exist
 (org.freedesktop.DBus.Error.UnknownObject))
```

I used QDBusViewer to look at the interface and it does seem to have that method:

```
<method name="WriteValue"><arg name="value" type="ay" direction="in"/>
<arg name="options" type="a{sv}" direction="in"/>
</method>
```

Any idea or suggestion on what I could be doing wrong? (I suspect that this is not really an issue in bluez-generated. But I thought I start here).

My code: https://github.com/ruediger/cloudbbq2-rs/blob/main/src/main.rs#L98

Thanks in advance.

(btw. would you be interested in moving bluetooth.rs into a separate crate? It seems very useful as a library. I'm happy to send a PR.)

qwandor:
Hi! I'm glad you've found it useful. The bluetooth.rs module is still a bit of a mess at the moment, and quite incomplete, so I'm not sure it's really in a state to be a separate crate yet, as that might make people think it's actually usable. The better approach long-term might be to move to something like [btleplug](https://crates.io/crates/btleplug) once it's more complete and stable. PRs are welcome though.

As for the WriteValue error, I've seen similar things the first time I try reading a characteristic after connecting to a device. I'm not entirely sure why; maybe it takes BlueZ a bit of time to expose everything over D-Bus. Does it work if you retry? You can try patching in the [retry branch](https://github.com/alsuren/mijia-homie/compare/retry) to see if it helps at all.

ruediger:
Thanks for the quick response! I've tried btleplug as well, but it either simply crashes or can't find devices. I've also tried https://crates.io/crates/bluez but unfortunately it's not very complete and although I can find devices, it lacks GATT functionality.

I'll try the the retry branch and report back.

alsuren:
[...]

PR #108, opened 2020-12-28
Moving bluetooth module of mijia crate out to new bluez-async crate.

qwandor:
As requested in #91. I changed my mind.


alsuren:
I really don't want to do this lightly. 

- [ ] We should at least talk to other rust bluetooth maintainers.

I've asked the btleplug maintainer (who also contributed the web-bluetooth bindings in web-sys) for their thoughts on API design. https://www.reddit.com/r/rust/comments/klvd4o/buttplugrs_hits_v1_milestone/ghdrgrg?utm_source=share&utm_medium=web2x&context=3 .

(I've not read your code yet. I will do the other PRs first)

qwandor:
I've commented on https://github.com/deviceplug/btleplug/issues/69 suggesting that I could implement a btleplug backend based on this crate. From having a look at the `btleplug` code, I think it wouldn't be too hard to do. I'd probably make a wrapper struct following the `btleplug` API but async, and then implement the API trait with a bunch of methods just blocking on the async versions. This would make it easy when `btleplug` moves to actually exposing an async API in future.

`dbus-tokio` just merged a PR to move to `tokio` 1.0, so I should be able to upgrade us as well once that makes it to a release. I'll just need to figure out how to use 0.2, 0.3 and 1.0 together the same binary (unless our other dependencies also upgrade before then).

For scope, my intention is to support (tokio-based) async, GAP/GATT central functionality (i.e. acting as a peripheral is out of scope), via BlueZ. I've added support for device-specific streams in #114.

bluez-async/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "bluez-async"
+version = "0.1.0"
+authors = ["Andrew Walbran <qwandor@google.com>", "David Laban <alsuren@gmail.com>"]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+description = "An async wrapper around the D-Bus interface of BlueZ, the Linux Bluetooth daemon."
```
qwandor, line 7: Explain the scope.

qwandor:
> * [x]  Think about scope
>   
>   * [x]  We currently only care about GATT properties, and we only care about reading/subscribing to an event stream at a global level. We also only care about being a client. If this is as far as we want to go, how do we communicate that this is the scope of our project?
Added notes to the README and description.

>   * [x]  if someone starts with our library and finds that they want to escape and do something non-gatt, can we provide them an escape hatch, so that they can use bluez-generated on raw proxy objects? (How do we make this feel sufficiently dirty that people don't do it when they should be using our in-scope APIs?)
They can convert the `DeviceId` or whatever to a dbus `Path`, and then construct a `Proxy` to use as they wish.

>   * [ ]  how are we avoiding the fate of blurz (everything hand coded to use sync dbus calls and would be painful to port to tokio, so the whole thing ended up unmaintained)?
Well, it's using tokio already at least!

>   * [x]  how are we avoiding the fate of bluez-dbus (interfaces for all of the possible concurrency models and never become confident enough to post on crates.io)?
I only intend to support async, and I am confident to post it on crates.io.

bluez-async/Cargo.toml
qwandor: Done.

bluez-async/README.md
```diff
@@ -0,0 +1,28 @@
...
+[![docs.rs page](https://docs.rs/bluez-async/badge.svg)](https://docs.rs/bluez-async)
+
+`bluez-async` is an async wrapper around the D-Bus interface of BlueZ, the Linux Bluetooth daemon.
+It provides type-safe interfaces to a subset of the Bluetooth client (i.e. central, in Bluetooth
+terminology) interfaces exposed by BlueZ, focussing on the Generic Attribute Profile (GATT) of
+Bluetooth Low Energy (BLE).
+
+See the [examples](examples/) directory for examples of how to use it.
+
+This crate is incomplete, experimental, and may not be supported.
```
alsuren, line 13: I was about to suggest `maintenance = { status = "experimental" }` but then I re-read https://doc.rust-lang.org/cargo/reference/manifest.html#the-badges-section and I guess this is just as good.

alsuren (approved):
Sorry I blocked this for so long.

PR #109, opened 2020-12-29
Cleanups to mijia crate and support for short UUIDs

qwandor:
Move all Bluetooth-related modules under `bluetooth`, clean up re-exports, fix keywords in `Cargo.toml`, and add support for converting UUIDs to and from 16- and 32-bit short forms.


mijia/src/bluetooth/bleuuid.rs
```diff
@@ -0,0 +1,108 @@
...
+
+/// Convert a 32-bit BLE short UUID to a full 128-bit UUID by filling in the standard Bluetooth Base
+/// UUID.
+pub const fn uuid_from_u32(short: u32) -> Uuid {
+    Uuid::from_u128(BLUETOOTH_BASE_UUID | ((short as u128) << 96))
+}
+
+/// Convert a 16-bit BLE short UUID to a full 128-bit UUID by filling in the standard Bluetooth Base
+/// UUID.
+pub const fn uuid_from_u16(short: u16) -> Uuid {
```
alsuren, line 15: Can these both be added to the BleUuid trait and called like `Uuid::from_u16(0x25)`? This should work a bit like how you implement Default::default() on structs.

qwandor: Unfortunately it doesn't seem to be possible to have const functions on a trait. )-:

alsuren (approved):

PR #111, opened 2020-12-30
Integration test for homie-controller talking to homie-device.

### LINKED ISSUE #52: Integration tests for `homie-device` and `homie-controller` together
Add a test to the `homie-device` crate which sets up a local MQTT broker, connects to it along with a Homie controller, and makes sure that the device advertised by `homie-device` is correctly picked up by `homie-controller`, and property values can be sent in both directions.

qwandor:
Starts a device and a controller connected to the same MQTT broker, and ensures that the controller can discover the device as expected.

Fixes #52.

### DIFF
--- a/homie-controller/tests/integration.rs
+++ b/homie-controller/tests/integration.rs
@@ -182,12 +182,20 @@ async fn test_device() {
     }
 
     // Disconnect the device.
-    homie.disconnect().await.unwrap();
-    let err = homie_handle.await.unwrap_err();
-    if let SpawnError::Connection(ConnectionError::MqttState(StateError::Io(e))) = err {
-        assert_eq!(e.kind(), ErrorKind::ConnectionAborted);
-    } else {
-        panic!("Unexpected error {:?}", err);
+    {
+        homie.disconnect().await.unwrap();
+        let err = homie_handle.await.unwrap_err();
+        if let SpawnError::Connection(ConnectionError::MqttState(StateError::Io(e))) = err {
+            assert_eq!(e.kind(), ErrorKind::ConnectionAborted);
+        } else {
+            panic!("Unexpected error {:?}", err);
+        }
+    }
+
+    // Disconnect the controller.
+    controller.disconnect().await.unwrap();
+    while let Ok(event) = controller.poll(&mut event_loop).await {
+        log::trace!("Event: {:?}", event);
     }
 }
 
--- a/homie-controller/tests/integration.rs
+++ b/homie-controller/tests/integration.rs
@@ -8,7 +8,7 @@ use homie_device::{HomieDevice, Node, Property, SpawnError};
 use librumqttd::{Broker, Config};
 use rumqttc::{ConnectionError, MqttOptions, StateError};
 use std::io::ErrorKind;
-use std::sync::{Arc, Mutex};
+use std::sync::mpsc;
 use std::thread;
 
 // A high port number which is hopefully not in use, to use for the MQTT broker.
@@ -25,14 +25,13 @@ async fn test_device() {
     controller.start().await.unwrap();
 
     // Start device
-    let updates = Arc::new(Mutex::new(vec![]));
-    let updates_ref = updates.clone();
+    let (updates_tx, updates_rx) = mpsc::sync_channel(10);
     let device_options = MqttOptions::new("homie_device", "localhost", PORT);
     let mut device_builder = HomieDevice::builder("homie/device_id", "Device name", device_options);
     device_builder.set_update_callback(move |node_id, property_id, value| {
         assert_eq!(property_id, "property_id");
         assert_eq!(node_id, "node_id");
-        updates_ref.lock().unwrap().push(value.clone());
+        updates_tx.send(value.clone()).unwrap();
         ready(Some(value))
     });
     let (mut homie, homie_handle) = device_builder.spawn().await.unwrap();
@@ -163,7 +162,10 @@ async fn test_device() {
             }
         }
     }
-    assert_eq!(*updates.lock().unwrap(), vec!["13".to_string()]);
+    assert_eq!(
+        updates_rx.try_iter().collect::<Vec<_>>(),
+        vec!["13".to_string()]
+    );
 
     // Check that the value sent back is reflected on the controller's view of the device.
     {
--- a/homie-controller/tests/integration.rs
+++ b/homie-controller/tests/integration.rs
@@ -162,10 +162,7 @@ async fn test_device() {
             }
         }
     }
-    assert_eq!(
-        updates_rx.try_iter().collect::<Vec<_>>(),
-        vec!["13".to_string()]
-    );
+    assert_eq!(updates_rx.try_iter().collect::<Vec<_>>(), vec!["13"]);
 
     // Check that the value sent back is reflected on the controller's view of the device.
     {
@@ -221,6 +218,9 @@ fn spawn_mqtt_broker(port: u16) {
     .unwrap();
     let mut broker = Broker::new(broker_config);
     thread::spawn(move || {
-        broker.start().unwrap();
+        broker.start().expect(&format!(
+            "Failed to start MQTT broker. This may be because port {} is already in use",
+            port,
+        ));
     });
 }
--- a/homie-controller/tests/integration.rs
+++ b/homie-controller/tests/integration.rs
@@ -2,11 +2,13 @@
 //! controller connected to the same MQTT broker, and ensures that the controller can discover the
 //! device as expected.
 
-use homie_controller::{HomieController, State};
+use futures::future::ready;
+use homie_controller::{Event, HomieController, State};
 use homie_device::{HomieDevice, Node, Property, SpawnError};
 use librumqttd::{Broker, Config};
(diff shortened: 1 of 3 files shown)


homie-controller/tests/integration.rs
```diff
@@ -0,0 +1,220 @@
...
+use homie_controller::{Event, HomieController, State};
+use homie_device::{HomieDevice, Node, Property, SpawnError};
+use librumqttd::{Broker, Config};
+use rumqttc::{ConnectionError, MqttOptions, StateError};
+use std::io::ErrorKind;
+use std::sync::{Arc, Mutex};
+use std::thread;
+
+// A high port number which is hopefully not in use, to use for the MQTT broker.
+const PORT: u16 = 10883;
```
alsuren, line 15: if you do `while cargo test -- homie-controller/tests/integration.rs ; do true; done` how many loops through do you get before it errors out?

homie-controller/tests/integration.rs
```diff
@@ -0,0 +1,220 @@
...
+async fn test_device() {
+    pretty_env_logger::init();
+
+    // Start MQTT broker.
+    spawn_mqtt_broker(PORT);
+
+    // Start controller.
+    let controller_options = MqttOptions::new("homie_controller", "localhost", PORT);
+    let (controller, mut event_loop) = HomieController::new(controller_options, "homie");
+    controller.start().await.unwrap();
```
alsuren, line 27: maybe make the function return Result and use ?

homie-controller/tests/integration.rs
```diff
@@ -0,0 +1,220 @@
...
+    port = 0
+    "#,
+        port
+    ))
+    .unwrap();
+    let mut broker = Broker::new(broker_config);
+    thread::spawn(move || {
+        broker.start().unwrap();
+    });
+}
```
alsuren, line 220: fancy fixing the clippy problems below while you're here?
|
|
V

homie-controller/tests/integration.rs
```diff
@@ -0,0 +1,220 @@
...
+    spawn_mqtt_broker(PORT);
+
+    // Start controller.
+    let controller_options = MqttOptions::new("homie_controller", "localhost", PORT);
+    let (controller, mut event_loop) = HomieController::new(controller_options, "homie");
+    controller.start().await.unwrap();
+
+    // Start device
+    let updates = Arc::new(Mutex::new(vec![]));
+    let updates_ref = updates.clone();
```
alsuren, line 31: you can use a channel for this

homie-controller/tests/integration.rs
```diff
@@ -0,0 +1,220 @@
...
+                assert_eq!(property_id, "property_id");
+                assert_eq!(value, "42");
+                assert_eq!(fresh, true);
+                break;
+            }
+        }
+    }
+
+    // Check that the device looks how we expect.
+    let devices = controller.devices();
```
alsuren, line 130: there is a lot of variable shadowing going on. Can we add some scopes to make the reader's brain work less hard?

homie-controller/tests/integration.rs
```diff
@@ -0,0 +1,220 @@
...
+    let err = homie_handle.await.unwrap_err();
+    if let SpawnError::Connection(ConnectionError::MqttState(StateError::Io(e))) = err {
+        assert_eq!(e.kind(), ErrorKind::ConnectionAborted);
+    } else {
+        panic!("Unexpected error {:?}", err);
+    }
+}
+
+/// Spawn an MQTT broker listening on the given port.
+fn spawn_mqtt_broker(port: u16) {
```
alsuren, line 185: This is very noisy. Can we set up some config/environment to make it shit up?

```
$ cargo test --test integration 
    Finished test [unoptimized + debuginfo] target(s) in 0.25s
     Running target/debug/deps/integration-62f3f2fd7256ed95

running 1 test
 ERROR librumqttd > Disconnected!! Id = 11, Client Id = homie_device
 ERROR librumqttd > Stopped!! Id = 12, Client Id = homie_controller, Network(Io(Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" }))
test test_device ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
```

homie-controller/tests/integration.rs
alsuren: ... `while cargo test --test integration ; do true; done` - seems to work fine forever.

qwandor: I guess that's fine then?

homie-controller/tests/integration.rs
qwandor: It works in some cases, but in a bunch of places I'm unwrapping `Option`s, so it doesn't work there. There are also a few inside closures. I think it's more consistent just to use `unwrap()` everywhere, and I guess will also give more useful backtraces.

homie-controller/tests/integration.rs
qwandor: Removing the `pretty_env_logger::init()` stops the error logs.

homie-controller/tests/integration.rs
qwandor: Done.

homie-controller/tests/integration.rs
qwandor: Done.

homie-controller/tests/integration.rs
qwandor: Will do in a different PR.

homie-controller/tests/integration.rs
alsuren: With logging enabled and running the loop in two different tabs eventually produces:
```
running 1 test
 ERROR librumqttd > Accept loop error: "I/O Address already in use (os error 48)"
test test_device ... FAILED

failures:

---- test_device stdout ----
thread 'test_device' panicked at 'called `Result::unwrap()` on an `Err` value: Request(SendError(..))', homie-controller/tests/integration.rs:53:32
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Without logging you might be a bit stuffed trying to debug that.

https://docs.rs/env_logger/0.8.2/env_logger/#capturing-logs-in-tests suggests a better way. Maybe pretty_env_logger has the same method?

homie-controller/tests/integration.rs
```diff
@@ -0,0 +1,226 @@
...
+    let (updates_tx, updates_rx) = mpsc::sync_channel(10);
+    let device_options = MqttOptions::new("homie_device", "localhost", PORT);
+    let mut device_builder = HomieDevice::builder("homie/device_id", "Device name", device_options);
+    device_builder.set_update_callback(move |node_id, property_id, value| {
+        assert_eq!(property_id, "property_id");
+        assert_eq!(node_id, "node_id");
+        updates_tx.send(value.clone()).unwrap();
+        ready(Some(value))
+    });
+    let (mut homie, homie_handle) = device_builder.spawn().await.unwrap();
```
alsuren, line 37: I guess you could also turn this into: 
```suggestion
    let (mut homie, homie_handle) = device_builder.spawn().await.expect("this may fail if port 10883 is already in use");
```

homie-controller/tests/integration.rs
```diff
@@ -0,0 +1,226 @@
...
+    controller
+        .set("device_id", "node_id", "property_id", 13)
+        .await
+        .unwrap();
+
+    // Wait for the device to receive the value and send it back to the controller.
+    loop {
+        if let Some(event) = controller.poll(&mut event_loop).await.unwrap() {
+            log::trace!("Event: {:?}", event);
+            if let Event::PropertyValueChanged {
```
alsuren, line 148: (not for today) I wonder if we want a helper function like `let event = controller.poll_until(|e| matches!(e, Event::PropertyValueChanged {..}))?` (vaguely equivalent to `s.skip_while(|e| !matches!(e, ..)).next()` for streams: https://docs.rs/futures/0.2.1/futures/stream/trait.StreamExt.html#method.skip_while)

homie-controller/tests/integration.rs
```diff
@@ -0,0 +1,226 @@
...
+                assert_eq!(property_id, "property_id");
+                assert_eq!(value, "13");
+                assert_eq!(fresh, true);
+                break;
+            }
+        }
+    }
+    assert_eq!(
+        updates_rx.try_iter().collect::<Vec<_>>(),
+        vec!["13".to_string()]
```
alsuren, line 167: I have a vague feeling that `vec![String::from("13")] == vec!["13"]` these days (or it will be soon?)

alsuren (approved):

homie-controller/tests/integration.rs
qwandor: I would expect it to be the `broker.start()` that fails, so I'll add an expect there.

homie-controller/tests/integration.rs
qwandor: Aha, it is! That's nicer.

homie-controller/tests/integration.rs
qwandor: Done. That means we get the error logs back, but I think the proper fix here is to avoid logging errors when nothing is actually wrong so I'll try to fix that.

PR #112, opened 2020-12-30
Add helper function to call introspect and parse XML.

qwandor:
This is a minor cleanup to put the introspection in the `introspect` module.

I tried to make an extension trait for `Introspectable`, but I couldn't see a way to do that without support for async functions on traits.

### DIFF
--- a/mijia/src/bluetooth/introspect.rs
+++ b/mijia/src/bluetooth/introspect.rs
@@ -1,5 +1,7 @@
 use dbus::nonblock::stdintf::org_freedesktop_dbus::Introspectable;
+use futures::Future;
 use serde_derive::Deserialize;
+use std::pin::Pin;
 
 use super::BluetoothError;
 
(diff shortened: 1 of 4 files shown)


mijia/src/bluetooth/introspect.rs
```diff
@@ -104,6 +107,12 @@ pub enum Access {
     Write,
 }
 
+pub async fn introspect(introspectable: &impl Introspectable) -> Result<Node, BluetoothError> {
```
alsuren, line 110: If you wanted to make this an extension trait, you just need concrete types (`Box<dyn Future>` rather than `impl Future`). https://github.com/dtolnay/async-trait is the best-practice way to do this.

I'm also happy to ship this as-is.

alsuren (approved):

mijia/src/bluetooth/introspect.rs
qwandor: Done.

PR #113, opened 2020-12-30
Rewrite Bluetooth event parsing.

qwandor:
This is a rewrite from scratch, using the proper types for parsing PropertiesChanged signals. It also:

 * Structures the events according to what type of object they are for.
 * Fixes a bug where only one event would be emitted when there should be several in the case that multiple properties change at once.
 * Adds some tests.
 * Removes the ServicesResolved event, as I can't see any obvious use-case for it at the moment.


mijia/src/bluetooth/events.rs
```diff
@@ -0,0 +1,205 @@
...
+        let mut changed_properties: HashMap<String, Variant<Box<dyn RefArg>>> = HashMap::new();
+        changed_properties.insert("RSSI".to_string(), Variant(Box::new(rssi)));
+        let properties_changed = PropertiesPropertiesChanged {
+            interface_name: "org.bluez.Device1".to_string(),
+            changed_properties,
+            invalidated_properties: vec![],
+        };
+        let message = properties_changed
+            .to_emit_message(&Path::new("/org/bluez/hci0/dev_A4_C1_38_39_2F_3E").unwrap());
+        let id = DeviceId::new("/org/bluez/hci0/dev_A4_C1_38_39_2F_3E");
```
qwandor, line 172: Don't use real MAC addresses in these tests.

mijia/src/bluetooth/bluetooth_event.rs
```diff
@@ -1,146 +0,0 @@
...
-        adapter: AdapterId,
-        discovering: bool,
-    },
-    Connected {
-        device: DeviceId,
-        connected: bool,
-    },
-    ServicesResolved {
-        device: DeviceId,
-        services_resolved: bool,
```
alsuren, line 50: Would ServicesResolved allow us to solve that bug when you connect() and then start_notify() and bluez says that it can't find the reading property's subscribe method and bombs out?

mijia/src/bluetooth/events.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// A new value of the characteristic has been received. This may be from a notification.
+    Value { value: Vec<u8> },
+}
+
+impl BluetoothEvent {
+    /// Return a `MatchRule` which will match all D-Bus messages which represent Bluetooth events.
+    pub(crate) fn match_rule() -> MatchRule<'static> {
+        // BusName validation just checks that the length and format is valid, so it should never
+        // fail for a constant that we know is valid.
+        let bus_name = "org.bluez".into();
```
alsuren, line 64: comment seems stale

alsuren (approved):
It's a thing of beauty :D

mijia/src/bluetooth/events.rs
qwandor: It's still valid. `into()` will panic if the string is invalid, but that won't happen for this constant string.

mijia/src/bluetooth/bluetooth_event.rs
qwandor: Possibly. But maybe that should be something internal to the library? It needs some more thought, in any case, and I don't think there's much value in exposing the event to client until we have a story for how it should be used.

PR #114, opened 2021-01-01
Add event for a new Bluetooth device being discovered

qwandor:
I've also done a bit of tidying up of the event parsing, added tests for the match rules, and added the ability to get a stream of events for a particular device or characteristic. This allows a TODO in `get_all_history` to be fixed.

### DIFF
--- a/mijia/src/bluetooth/events.rs
+++ b/mijia/src/bluetooth/events.rs
@@ -133,7 +133,6 @@ mod tests {
     use std::collections::HashMap;
 
     use dbus::arg::{RefArg, Variant};
-    use dbus::Path;
 
     use super::*;
 
@@ -146,7 +145,7 @@ mod tests {
             changed_properties,
             invalidated_properties: vec![],
         };
-        let message = properties_changed.to_emit_message(&Path::new("/org/bluez/hci0").unwrap());
+        let message = properties_changed.to_emit_message(&"/org/bluez/hci0".into());
         let id = AdapterId::new("/org/bluez/hci0");
         assert_eq!(
             BluetoothEvent::message_to_events(message),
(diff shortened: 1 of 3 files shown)


mijia/src/bluetooth/events.rs
```diff
@@ -57,96 +61,144 @@ pub enum CharacteristicEvent {
...
+
+        let mut match_rules = vec![];
+
+        // If we aren't filtering to a single device or characteristic, then match ObjectManager
+        // signals so we can get events for new devices being discovered.
+        if object.is_none() {
+            let match_rule =
+                ObjectManagerInterfacesAdded::match_rule(Some(&bus_name), None).static_clone();
+            match_rules.push(match_rule);
+        }
```
alsuren, line 83: (optional) I was already a bit skeptical about match_rule() living inside BluetoothEvent (but it is not public api so I didn't say anything). I think I would split this into a couple of functions in the same module as the caller (just `filtered_event_stream()` currently), and make the caller responsible for calling this as a separate function (but it is not public api so I'm also okay with it as it is).

mijia/src/bluetooth/events.rs
```diff
@@ -57,96 +61,144 @@ pub enum CharacteristicEvent {
...
+        // under it. If no object is specified then this will match PropertiesChanged signals for
+        // all BlueZ objects.
+        {
+            let object_path = object.map(|o| o.into());
+            let mut match_rule =
+                PropertiesPropertiesChanged::match_rule(Some(&bus_name), object_path.as_ref())
+                    .static_clone();
+            match_rule.path_is_namespace = true;
+            match_rules.push(match_rule);
+        }
```
alsuren, line 95: scope feels unnecessary

alsuren (approved):

mijia/src/bluetooth/events.rs
qwandor: I quite like having this logic in the same module as `message_to_events`, because it kind of parallels it. I think of this module as having as much as possible of the logic for dealing with events without having anything async or actually talking to D-Bus, which means it can easily be tested.

mijia/src/bluetooth/events.rs
qwandor: Removed. I think I added because you made a comment in a previous code review about liking scopes for something similar to this.

alsuren: That was mostly me desperately trying to find a way to make sense of a 200-line test function. Splitting into separate functions would always be preferable for real code.

PR #116, opened 2021-01-03
Add support for descriptors

qwandor:
Also clean up characteristic example a bit, and implement `Display` for ID types.

### DIFF
--- a/mijia-homie/src/main.rs
+++ b/mijia-homie/src/main.rs
@@ -414,8 +414,8 @@ async fn connect_and_subscribe_sensor_or_disconnect(
             .bt_session
             .disconnect(&id)
             .await
-            .wrap_err_with(|| format!("Disconnecting from {} ({:?})", name, id))?;
-        Err(Report::new(e).wrap_err(format!("Starting notifications on {} ({:?})", name, id)))
+            .wrap_err_with(|| format!("Disconnecting from {} ({})", name, id))?;
+        Err(Report::new(e).wrap_err(format!("Starting notifications on {} ({})", name, id)))
     })
     .await?;
 
@@ -447,7 +447,7 @@ async fn check_for_stale_sensor(
             .bt_session
             .disconnect(id)
             .await
-            .wrap_err_with(|| format!("disconnecting from {:?}", id))?;
+            .wrap_err_with(|| format!("disconnecting from {}", id))?;
     }
     Ok(())
 }
@@ -486,7 +486,7 @@ async fn handle_bluetooth_event(
                     ConnectionStatus::Connected { id: connected_id } => {
                         if id != *connected_id {
                             log::info!(
-                                "Got update from device on unexpected id {:?} (expected {:?})",
+                                "Got update from device on unexpected id {} (expected {})",
                                 id,
                                 connected_id,
                             );
@@ -494,13 +494,13 @@ async fn handle_bluetooth_event(
                     }
                     ConnectionStatus::Connecting { .. } => {}
                     _ => {
-                        println!("Got update from disconnected device {:?}. Connecting.", id);
+                        println!("Got update from disconnected device {}. Connecting.", id);
                         sensor.mark_connected(homie, id).await?;
                         // TODO: Make sure the connection interval is set.
                     }
                 }
             } else {
-                println!("Got update from unknown device {:?}.", id);
+                println!("Got update from unknown device {}.", id);
             }
         }
         MijiaEvent::Disconnected { id } => {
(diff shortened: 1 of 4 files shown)


mijia-homie/src/main.rs
```diff
@@ -513,18 +513,18 @@ async fn handle_bluetooth_event(
...
                     }
                 } else {
                     println!(
-                        "{} ({:?}) disconnected but wasn't known to be connected.",
+                        "{} ({}) disconnected but wasn't known to be connected.",
                         sensor.name, id
                     );
                 }
             } else {
-                println!("Unknown device {:?} disconnected.", id);
```
alsuren, line 527: let me guess: the debug representation of Path is mental?

mijia/examples/characteristics.rs
```diff
@@ -1,37 +1,78 @@
...
+                    }
+                }
             }
         }
     }
 
     Ok(())
 }
+
+const PRINTABLE_ASCII_RANGE: RangeInclusive<u8> = 0x20..=0x7E;
```
alsuren, line 55: does this work?

```suggestion
const PRINTABLE_ASCII_RANGE: RangeInclusive<u8> = ' '..='~';
```

mijia/examples/characteristics.rs
```diff
@@ -1,37 +1,78 @@
...
+                    } else {
+                        "secondary"
+                    },
+                    service.id
                 );
+                let characteristics = session.get_characteristics(&service.id).await?;
+                for characteristic in characteristics {
+                    println!(
+                        "  Characteristic {}: {}",
+                        characteristic.uuid.to_short_string(),
```
alsuren, line 34:
```suggestion
                        characteristic.uuid.succinctly(),
```
?

alsuren: does kinda feel like we might want to newtype uuid and control its display impl though, rather than using a trait.

mijia/src/bluetooth/mod.rs
```diff
@@ -406,6 +511,34 @@ impl BluetoothSession {
...
+    ) -> Result<Vec<DescriptorInfo>, BluetoothError> {
+        let characteristic_node = self
+            .characteristic(characteristic)
+            .introspect_parse()
+            .compat()
+            .await?;
+        let mut descriptors = vec![];
+        for subnode in characteristic_node.nodes {
+            let subnode_name = subnode.name.as_ref().unwrap();
+            if subnode_name.starts_with("desc") {
```
alsuren, line 527: is this rule documented anywhere or should we *strictly* be introspecting each path and only returning things that have the appropriate dbus interface on them?

(I'm fine with this staying as is, but it probably wants a comment on it)

mijia/src/bluetooth/mod.rs
```diff
@@ -406,6 +511,34 @@ impl BluetoothSession {
...
+                let descriptor_id = DescriptorId {
+                    object_path: format!("{}/{}", characteristic.object_path, subnode_name).into(),
+                };
+                let uuid =
+                    Uuid::parse_str(&self.descriptor(&descriptor_id).uuid().compat().await?)?;
+                descriptors.push(DescriptorInfo {
```
alsuren, line 533: What does this for loop look like as a .filter_map().collect()?

alsuren (approved):

mijia-homie/src/main.rs
qwandor: It's not terrible, but something like DeviceId { Path { "/org/bluez/hci0/dev_11_22_33_44_55_66" } } seems unhelpfully verbose for a log message, as a user of the service doesn't really care how we represent things internally. The adapter and MAC address are the parts which are likely to be useful, so it seems nicer to just print that.

mijia/examples/characteristics.rs
qwandor: No, because those literals are `char`s not `u8`s.

mijia/src/bluetooth/mod.rs
qwandor: Painful, because of the async function call and error handling.

mijia/src/bluetooth/mod.rs
qwandor: It is documented incorrectly in https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/gatt-api.txt that descriptors should have paths of the form `[variable prefix]/{hci0,hci1,...}/dev_XX_XX_XX_XX_XX_XX/serviceXX/charYYYY/descriptorZZZ`. The last part of the path is actually of the form `serviceXXXX/charYYYY/descZZZZ`.

Added a comment.

alsuren: An opportunity for your first bluez patch ;-)

mijia/examples/characteristics.rs
qwandor: Renamed. I think I still want to keep UUID as the type in the API.

PR #118, opened 2021-01-03
Add flags for characteristic.

qwandor:
These describe what operations the characteristic supports, e.g. reading, writing, notifying.

### DIFF
--- a/mijia/src/bluetooth/mod.rs
+++ b/mijia/src/bluetooth/mod.rs
@@ -7,6 +7,7 @@ pub use self::bleuuid::{uuid_from_u16, uuid_from_u32, BleUuid};
 pub use self::events::{AdapterEvent, BluetoothEvent, CharacteristicEvent, DeviceEvent};
 use self::introspect::IntrospectParse;
 use self::messagestream::MessageStream;
+use bitflags::bitflags;
 use bluez_generated::{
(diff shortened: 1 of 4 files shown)


mijia/src/bluetooth/mod.rs
```diff
@@ -329,6 +334,61 @@ pub struct CharacteristicInfo {
...
+        const INDICATE = 0x20;
+        const SIGNED_WRITE = 0x40;
+        const EXTENDED_PROPERTIES = 0x80;
+        const RELIABLE_WRITE = 0x100;
+        const WRITABLE_AUXILIARIES = 0x200;
+        const ENCRYPT_READ = 0x400;
+        const ENCRYPT_WRITE = 0x800;
+        const ENCRYPT_AUTHENTICATED_READ = 0x1000;
+        const ENCRYPT_AUTHENTICATED_WRITE = 0x2000;
+        const AUTHORIZE = 0x4000;
```
alsuren, line 360: do these values happen to be the same ones that the bluetooth spec uses?

alsuren (approved):

mijia/src/bluetooth/mod.rs
qwandor: The first 10 are at least. Or really, the first 8 are what goes in the properties field, `RELIABLE_WRITE` and `WRITABLE_AUXILIARIES` go in the extended properties descriptor. The others don't seem to be mentioned in the spec as far as I can see, but are listed in the BlueZ documentation so I figured we should probably include them.

PR #120, opened 2021-01-06
Use latest dbus-codegen

### LINKED ISSUE #9: Generate strongly typed deserialiser for get_managed_objects() variants.

qwandor:
This includes my addition https://github.com/diwic/dbus-rs/pull/304 to add type-safe property parsing, which makes things a bit neater.

Fixes #9.


bluez-generated/src/leadvertisingmanager1.rs
```diff
@@ -62,3 +62,28 @@ impl<'a, T: nonblock::NonblockReply, C: ::std::ops::Deref<Target = T>> OrgBluezL
...
+    }
+
+    pub fn supported_instances(&self) -> Option<u8> {
+        arg::prop_cast(self.0, "SupportedInstances").copied()
+    }
+
+    pub fn supported_includes(&self) -> Option<&Vec<String>> {
+        arg::prop_cast(self.0, "SupportedIncludes")
+    }
+}
```
alsuren, line 89: I feel like this PR validates our earlier decision to make a bluez-generated crate that *only* contains generated code.

mijia/src/bluetooth/mod.rs
```diff
@@ -481,30 +480,16 @@ impl BluetoothSession {
...
-                    .to_string();
-                let name = device_properties.get("Name").map(|name| {
-                    name.as_iter()
-                        .unwrap()
-                        .filter_map(|addr| addr.as_str())
-                        .next()
-                        .unwrap()
-                        .to_string()
-                });
+                let device_properties = OrgBluezDevice1Properties::from_interfaces(&interfaces)?;
```
alsuren, line 483: (mostly rambling and probably re-hashing the discussion from your upstream PR, so feel free to ignore me)

I feel like it's a bit surprising that device_properties owns a reference into `interfaces`, but I guess it is efficient and device_properties will typically be short-lived, so it's probably fine (I also can't think of a better way to do it).

I suppose if you were only asking for a single property then you'd be using a fluent interface (`OrgBluezDevice1Properties::from_interfaces(&interfaces)?.address()?`) and it would not be surprising at all.

I suppose if OrgBluezDevice1Properties was a trait with a blanket impl over `&'a ::std::collections::HashMap<String, arg::PropMap>` then it could be written `OrgBluezDevice1Properties::address(&interfaces)`, but people would be tempted to `use OrgBluezDevice1Properties` and write `interfaces.address()?` which might be a bit fragile.

mijia/src/bluetooth/mod.rs
```diff
@@ -823,32 +808,26 @@ impl BluetoothSession {
...
-                    .as_iter()?
-                    .map(|el| Some(el.as_u64()? as u8))
-                    .collect();
-                let v = v?;
-                Some((k, v))
+            .service_data()?
+            .iter()
+            .filter_map(|(k, v)| match Uuid::parse_str(k) {
+                Ok(uuid) => {
+                    let v = cast::<Vec<u8>>(&v.0)?;
```
alsuren, line 824: (optional) should we be noisier if this cast fails?

alsuren (approved):

bluez-generated/src/leadvertisingmanager1.rs
qwandor: Generating it as part of a `build.rs` would also work, but this is simpler in some ways.

mijia/src/bluetooth/mod.rs
qwandor: Added a warning log.

mijia/src/bluetooth/mod.rs
qwandor: Yeah, it looks a little weird, but seems to work well in practice so far. The other issue with the trait approach is that I want `from_interfaces` to return something that can be passed around, which is a bit of a pain if it's a trait object.

Note that `from_intefaces` is doing one level of HashMap lookup, and `address()` is doing a second level.

PR #124, opened 2021-01-12
Clean up and expand documentation

qwandor:
* Improve links in the top-level README.
 * Wrap Markdown files consistently at 100 columns.
 * Add examples to README for `mijia` and `bluez-async`.
 * Some other minor fixes.


bluez-generated/README.md
```diff
@@ -3,26 +3,39 @@
...
-Only async bindings are generated. Blocking bindings could also be generated,
-but I'm unlikely to use them, so they would need to be contributed by someone
-else.
+Only async bindings are generated. Blocking bindings could also be generated, but I'm unlikely
+to use them, so they would need to be contributed by someone else.
 
-It would be nice to generate some strongly typed bindings around
-`get_managed_objects()` for bulk-fetching properties and `add_match()` for
-subscribing to events.
+It would be nice to generate some strongly typed bindings around `add_match()` for subscribing to
```
qwandor, line 25: What does this mean actually?

bluez-generated/README.md
```diff
@@ -3,10 +3,11 @@
 [![crates.io page](https://img.shields.io/crates/v/bluez-generated.svg)](https://crates.io/crates/bluez-generated)
 [![docs.rs page](https://docs.rs/bluez-generated/badge.svg)](https://docs.rs/bluez-generated)
 
-This crate contains async bindings for Bluez.
+This crate contains async D-Bus bindings for BlueZ.
```
alsuren, line 6: Maybe the same treatment for the description in Cargo.toml ?

(not for now, but something that I've been pondering):

It's also worth noting that crates.io loads the crate page in two parts, so if you are on a slow connection with no cache, you will notice that it flashes up a page with just the description and then fills in the README.md afterwards. This is a bit janky.

In the back of my mind, I think it might be possible to reduce the jank by repeating the description in README.md **exactly**, but I've not tried it yet. I also don't know how it would interact with badges)

bluez-generated/README.md
alsuren: Let's say we have an interface with a `connected` property which is known to be boolean. It would be interesting to generate a `.connected_changes() -> Stream<Item=bool>` that listens for property-changed signals and presents them as a stream.

mijia/README.md
```diff
@@ -3,9 +3,51 @@
...
 
-See the [examples](examples/) directory for examples of how to use it.
+Currently only supports running on Linux, as it depends on BlueZ for Bluetooth.
+
+## Usage
+
+```rust
+// Create a new session. This establishes the D-Bus connection. In this case we ignore the join
+// handle, as we don't intend to run indefinitely.
+let (_, session) = MijiaSession::new().await?;
```
alsuren, line 15: bonus points if this can be tested using something like https://github.com/budziq/rust-skeptic :D

alsuren (approved):

mijia/README.md
alsuren: (skeptic itself recommends using https://crates.io/crates/doc-comment for this use-case)

bluez-generated/README.md
qwandor: Okay, I've clarified that. Though I think the match would just have to be on the `PropertiesChanged` signal, which would include all properties of the object, so it might make sense to return a stream of the *Properties structs that dbus-codegen now generates.

mijia/README.md
qwandor: Unfortunately I can't get `doc-comment` to work, because it doesn't seem to have any way to add the surrounding boilerplate to let `.await` and `?` work in the example. `rust-skeptic` might work, but requires adding a `build.rs` file which I'd prefer not to.

bluez-generated/README.md
qwandor: Edited to make the first sentence the same as in Cargo.toml.

PR #126, opened 2021-01-13
Increment version numbers for releasing bluez-async and mijia.

qwandor:
This lets us get the improved README onto crates.io, as well as the extra fields on `DeviceInfo`.


alsuren (approved):
Seems reasonable. You can probably merge version bumps straight to master without approval.

Interesting clippy lints from a previous pr. I agree with one of them but not sure about the other.

PR #127, opened 2021-01-13
Go back to building cross Docker images from Debian.

### LINKED ISSUE #125: Debian packages link against the wrong version of libssl
The package for homie-influx 0.2.1 apparently links against libssl 1.0.2, but Debian stable has 1.0.0. This prevents it from running.

```
$ ldd -d /usr/bin/homie-influx 
/usr/bin/homie-influx: /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.0: version `OPENSSL_1.0.2' not found (required by /usr/bin/homie-influx)
/usr/bin/homie-influx: /usr/lib/x86_64-linux-gnu/libssl.so.1.0.0: version `OPENSSL_1.0.2' not found (required by /usr/bin/homie-influx)
        linux-vdso.so.1 (0x00007ffeed3fb000)
        libssl.so.1.0.0 => /usr/lib/x86_64-linux-gnu/libssl.so.1.0.0 (0x00007ff4302b2000)
        libcrypto.so.1.0.0 => /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.0 (0x00007ff42feb5000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007ff42fe9b000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007ff42fe91000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007ff42fe70000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff42fced000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007ff42fce6000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff42fb25000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff430eb0000)
symbol X509_VERIFY_PARAM_set1_host version OPENSSL_1.0.2 not defined in file libcrypto.so.1.0.0 with link time reference        (/usr/bin/homie-influx)
symbol X509_VERIFY_PARAM_set_hostflags version OPENSSL_1.0.2 not defined in file libcrypto.so.1.0.0 with link time reference    (/usr/bin/homie-influx)
symbol X509_VERIFY_PARAM_set1_ip version OPENSSL_1.0.2 not defined in file libcrypto.so.1.0.0 with link time reference  (/usr/bin/homie-influx)
symbol SSL_get0_param version OPENSSL_1.0.2 not defined in file libssl.so.1.0.0 with link time reference        (/usr/bin/homie-influx)

```

qwandor:
Using an image based on the upstream cross images doesn't really work because they are based on an old version of Ubuntu, so we end up linking against the wrong version of libssl.

Fixes #125, by essentially reverting #102.

### DIFF
--- a/docker/Dockerfile.debian-buster-aarch64
+++ b/docker/Dockerfile.debian-buster-aarch64
@@ -1,5 +1,5 @@
 # Built from https://github.com/qwandor/cross/blob/context/docker/Dockerfile.context
-FROM ghcr.io/qwandor/cross-context:latest as context
+FROM ghcr.io/qwandor/cross-context:0.2.1 as context
 
 FROM debian:buster
 
--- a/docker/Dockerfile.debian-buster-aarch64
+++ b/docker/Dockerfile.debian-buster-aarch64
@@ -1,3 +1,4 @@
+# Built from https://github.com/qwandor/cross/blob/context/docker/Dockerfile.context
 FROM ghcr.io/qwandor/cross-context:latest as context
 
 FROM debian:buster
--- a/docker/Dockerfile.debian-buster-aarch64
+++ b/docker/Dockerfile.debian-buster-aarch64
@@ -1,0 +1,27 @@
+FROM ghcr.io/qwandor/cross-context:latest as context
+
+FROM debian:buster
+
+COPY --from=context common.sh lib.sh /
+RUN /common.sh
+
+COPY --from=context cmake.sh /
+RUN /cmake.sh
+
+COPY --from=context xargo.sh /
+RUN /xargo.sh
+
+RUN apt-get install --assume-yes --no-install-recommends \
+	g++-aarch64-linux-gnu \
+	libc6-dev-arm64-cross
+
+RUN dpkg --add-architecture arm64 && \
+	apt-get update && \
+	apt-get install -y libssl-dev:arm64 libdbus-1-dev:arm64
+
+ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc \
+	CC_aarch64_unknown_linux_gnu=aarch64-linux-gnu-gcc \
+	CXX_aarch64_unknown_linux_gnu=aarch64-linux-gnu-g++ \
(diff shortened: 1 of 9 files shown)


docker/Dockerfile.debian-buster-aarch64
```diff
@@ -0,0 +1,28 @@
+# Built from https://github.com/qwandor/cross/blob/context/docker/Dockerfile.context
+FROM ghcr.io/qwandor/cross-context:latest as context
```
alsuren, line 2: maybe tag this as a particular version (whatever the current version or cross is?) to avoid the problem that I caused when I updated my branch and rebuilt?

docker/update.sh
```diff
@@ -0,0 +1,14 @@
...
+set -euo pipefail
+
+docker build ./docker -f docker/Dockerfile.debian-buster-aarch64 \
+    -t ghcr.io/qwandor/cross-dbus-debian-buster-aarch64:latest
+docker build ./docker -f docker/Dockerfile.debian-buster-armv7 \
+    -t ghcr.io/qwandor/cross-dbus-debian-buster-armv7:latest
+docker build ./docker -f docker/Dockerfile.debian-buster-x86_64 \
+    -t ghcr.io/qwandor/cross-dbus-debian-buster-x86_64:latest
+
+docker push ghcr.io/qwandor/cross-dbus-debian-buster-aarch64:latest
```
alsuren, line 12: maybe also tag this with a version number (defined as a constant at the top of the file?) and push that too? Then you can use this versioned image in Cross.toml and have a better chance at avoiding breaking the build later

alsuren (approved):
I can believe that this will work fine. Some suggestions to avoid getting into the mess that I got us in before. Can all happen later though.

docker/Dockerfile.debian-buster-aarch64
qwandor: Done.

docker/update.sh
qwandor: Done.

Issue #129, opened 2021-01-21
Make manufacturer specific data available


MattesWhite:
First of all thank you for this interesting BLE library.

From the docs I don't see the possibility to access the manufacturer specific data advertised by some BLE devices. Would it be possible to expose them, maybe as an event?

Furthermore, it would be nice to be able to configure the scan process. Especially, I have the `filter duplicates` option in mind.

---

### Background

I'm using [Ruuvi Tags](https://ruuvi.com/ruuvitag-specs/) in my application. They use the manufacturer specific data advertisements to broadcast their sensor data. To be able to access the sensor data, I need access to the manufacturer specific data and be able to turn off the duplicates filter as otherwise newer values wouldn't be recognized.

qwandor:
Sounds reasonable. It should be fairly straightforward to add a manufacturer data field to the `DeviceInfo` struct. We should probably also add a `BluetoothSession::get_device_info` method, similar to the existing `get_service_info` and so on, to get a single device rather than having to get them all with `get_devices`.

It looks like scan configuration is done by [calling `SetDiscoveryFilter` on the adapter](https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/adapter-api.txt). This could be done by `BluetoothSession::start_discovery` or by a new method. We should probably also add a `stop_discovery` method as we don't yet have that.

Feel free to send a pull request, or else I can have a look tonight or this weekend.

qwandor:
Oh, and it sounds like you also want a new `DeviceEvent` variant. I haven't looked into that yet.

qwandor:
@MattesWhite Please take a look at #130 and see if it does what you want, and what you think of the code structure. If you give me a code review on that then I'll merge it and release a new version for you.

MattesWhite:
Thanks a lot for the big amount of work you put into my request. I really didn't expected that much. As requested I reviewed the PR, don't have anything to complain about.

To finish things, I'll test the events example with my setup to acknowledge that everything works.

qwandor:
Fixed by #130.

PR #130, opened 2021-01-21
Add manufacturer data to DeviceInfo, and add new DeviceEvent variant

qwandor:
This does part of what is requested in #129.

Also adds new `get_device_info` and `stop_discovery` methods.

### DIFF
--- a/bluez-async/src/events.rs
+++ b/bluez-async/src/events.rs
@@ -7,8 +7,9 @@ use dbus::nonblock::stdintf::org_freedesktop_dbus::{
     ObjectManagerInterfacesAdded, PropertiesPropertiesChanged,
 };
 use dbus::{Message, Path};
+use std::collections::HashMap;
 
-use super::{AdapterId, CharacteristicId, DeviceId};
+use super::{convert_manufacturer_data, AdapterId, CharacteristicId, DeviceId};
 
 /// An event relating to a Bluetooth device or adapter.
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -54,6 +55,10 @@ pub enum DeviceEvent {
     Connected { connected: bool },
     /// A new value is available for the RSSI of the device.
     RSSI { rssi: i16 },
+    /// A new value is available for the manufacturer-specific advertisement data of the device.
+    ManufacturerData {
+        manufacturer_data: HashMap<u16, Vec<u8>>,
+    },
 }
 
 /// Details of an event related to a GATT characteristic.
@@ -171,9 +176,17 @@ impl BluetoothEvent {
                 }
                 if let Some(rssi) = device.rssi() {
                     events.push(BluetoothEvent::Device {
-                        id,
+                        id: id.clone(),
                         event: DeviceEvent::RSSI { rssi },
                     });
+                }
+                if let Some(manufacturer_data) = device.manufacturer_data() {
+                    events.push(BluetoothEvent::Device {
+                        id,
+                        event: DeviceEvent::ManufacturerData {
+                            manufacturer_data: convert_manufacturer_data(manufacturer_data),
+                        },
+                    })
                 }
             }
             ORG_BLUEZ_GATT_CHARACTERISTIC1_NAME => {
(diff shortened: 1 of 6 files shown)


bluez-async/src/events.rs
```diff
@@ -47,17 +49,23 @@ pub enum AdapterEvent {
 
 /// Details of an event related to a Bluetooth device.
 #[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
```
alsuren, line 52: I suppose this is an advantage of the cargo-release workflow: if master always has a -alpha version number then it can be bumped as part of any PRs that make breaking changes.

bluez-async/src/events.rs
```diff
@@ -47,17 +49,23 @@ pub enum AdapterEvent {
...
     RSSI { rssi: i16 },
+    /// A new value is available for the manufacturer-specific advertisement data of the device.
+    ManufacturerData {
+        manufacturer_data: HashMap<u16, Vec<u8>>,
+    },
 }
 
 /// Details of an event related to a GATT characteristic.
 #[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
```
alsuren, line 68: what are we expecting to add in the future?

bluez-async/src/lib.rs
```diff
@@ -493,6 +618,42 @@ impl BluetoothSession {
...
+
+    /// Get a list of all Bluetooth adapters on the system.
+    async fn get_adapters(&self) -> Result<Vec<AdapterId>, dbus::Error> {
+        let bluez_root = Proxy::new(
+            "org.bluez",
+            "/",
+            DBUS_METHOD_CALL_TIMEOUT,
+            self.connection.clone(),
+        );
+        // TODO: See whether there is a way to do this with introspection instead, rather than
```
alsuren, line 644: add a ticket for this?

mijia/src/lib.rs
```diff
@@ -158,7 +158,7 @@ impl MijiaEvent {
 /// let (_, session) = MijiaSession::new().await?;
 ///
 /// // Start scanning for Bluetooth devices, and wait a few seconds for some to be discovered.
-/// session.bt_session.start_discovery().await?;
+/// session.bt_session.start_discovery(&Default::default()).await?;
```
alsuren, line 161: do we have anything in a README.md that needs changing?

alsuren (approved):
🔍

mijia/src/lib.rs
qwandor: I updated this in two README.md files already, is there anything else I missed?

bluez-async/src/events.rs
qwandor: I don't have anything particularly in mind, but it seems like a good idea in general to avoid more breaking changes.

bluez-async/src/lib.rs
qwandor: #135

bluez-async/src/lib.rs
```diff
@@ -458,33 +586,30 @@ impl BluetoothSession {
...
+    /// devices.
+    ///
+    /// Note that BlueZ combines discovery filters from all clients and sends events matching any
+    /// filter to all clients, so you may receive unexpected discovery events if there are other
+    /// clients on the system using Bluetooth as well.
+    ///
+    /// In most common cases, `DiscoveryFilter::default()` is fine.
+    pub async fn start_discovery(
+        &self,
+        discovery_filter: &DiscoveryFilter,
```
MattesWhite, line 599: As you already stated in the comment, in most cases the default filter is fine. Therefore, I suggest to keep the old `start_discovery()` but add a new method `start_discovery_with_filter()`. Then `start_discovery()` can look like this:

```rust
pub async fn start_discovery(&self) -> Result<(), BluetoothError> {
    self.start_discovery_with_filter(&Default::default)
}
```

This way you don't have to break the API or change the README.md .

MattesWhite (commented):
Nice work :+1:

bluez-async/src/lib.rs
qwandor: There's already a breaking change because of adding enum variants, but done.

qwandor:
Thanks for the reviews, merging.

PR #132, opened 2021-01-23
Split out separate modules to make lib.rs shorter.

qwandor:
No functional changes, this is purely moving code around.


alsuren (approved):
🍴

PR #133, opened 2021-01-23
Pass through device discovery events.

qwandor:
This might be useful to make `mijia-homie` more event driven.

### DIFF
--- a/mijia/src/lib.rs
+++ b/mijia/src/lib.rs
@@ -6,6 +6,11 @@
 //!
 //! [`MijiaSession']: struct.MijiaSession.html
 
+pub use bluez_async as bluetooth;
+use bluez_async::{
+    BluetoothError, BluetoothEvent, BluetoothSession, CharacteristicEvent, DeviceEvent, DeviceId,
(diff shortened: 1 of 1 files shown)


mijia/src/lib.rs
```diff
@@ -138,6 +140,22 @@ impl MijiaEvent {
...
+                id,
+                event: DeviceEvent::Discovered,
+            } => {
+                // Only pass through discovery events for sensors we recognise.
+                let device = session
+                    .get_device_info(&id)
+                    .await
+                    .map_err(|e| log::error!("Error getting device info: {:?}", e))
+                    .ok()?;
+                if device.name.as_deref() == Some(MIJIA_NAME) {
```
alsuren, line 153: (optional) I might be tempted to write a `is_mijia(device)` helper in case we manage to get the service property method working reliably again in the future.

alsuren (approved):

mijia/src/lib.rs
qwandor: Done.

PR #134, opened 2021-01-24
Helper example to add names to a config file

alsuren:
Makes the assumption that the config file only contains "name" = "value" entries, and no tables. This matches what we're currently doing in mijia-homie. I think it should bail out early if this assumption doesn't hold, but I haven't tested it.

This is what I used to create `pi@cottagepi.local:mijia-homie/sensor-names.toml` (with a bit of iteration), which includes the addresses of everything in the tower of sensors in the living room, and a bunch of addresses for sensors that I didn't manage to find.

I probably should have made a PR out of this as soon as I made it.

Open questions:
- [x] Is `mijia` the best place for this, or is it more `mijia-homie` specific?
- [ ] I guess I should test it properly by promoting pi@cottagepi.local:mijia-homie/sensor-names.toml to be the one that we're actually using?

### DIFF
--- a/mijia-homie/src/mijia-names.rs
+++ b/mijia-homie/src/mijia-names.rs
@@ -87,6 +87,12 @@ async fn main() -> Result<(), Report> {
         }
     }
 
+    println!(
+        "Finished inspecting all known sensors.\n\n\
+        Some sensors may have been discovered while we were trying to inspect \
+        this batch. Re-run the program to also inspect these new sensors."
+    );
+
     Ok(())
 }
 
--- a/mijia-homie/src/mijia-names.rs
+++ b/mijia-homie/src/mijia-names.rs
@@ -57,6 +57,10 @@ async fn main() -> Result<(), Report> {
             let sensor_names_filename = sensor_names_filename.clone();
 
             tokio::task::spawn_blocking(move || {
+                println!(
+                    "Successfully connected to {} (it should now have a bluetooth icon on it)",
+                    sensor.mac_address
+                );
                 println!("What name do you want to use for {}?", sensor.mac_address);
                 let mut name = String::new();
                 stdin().read_line(&mut name)?;
--- a/mijia-homie/src/mijia-names.rs
+++ b/mijia-homie/src/mijia-names.rs
@@ -42,7 +42,8 @@ async fn main() -> Result<(), Report> {
     {
         println!("Connecting to {}", sensor.mac_address);
         if let Err(e) = session.bt_session.connect(&sensor.id).await {
-            println!("Failed to connect to {}: {:?}", sensor.mac_address, e);
+            println!("Failed to connect to {}", sensor.mac_address);
+            log::debug!("error was: {:?}", e);
 
             let mut file = OpenOptions::new()
                 .create(true)
--- a/mijia-homie/src/mijia-names.rs
+++ b/mijia-homie/src/mijia-names.rs
@@ -1,6 +1,6 @@
 //! Example of how to subscribe to readings from one or more sensors.
 
-use bluez_async::MacAddress;
+use mijia::bluetooth::MacAddress;
 use eyre::Report;
 use eyre::WrapErr;
 use mijia::{MijiaSession, SensorProps};
--- a/mijia-homie/src/mijia-names.rs
+++ b/mijia-homie/src/mijia-names.rs
@@ -1,8 +1,8 @@
 //! Example of how to subscribe to readings from one or more sensors.
 
-use mijia::bluetooth::MacAddress;
 use eyre::Report;
 use eyre::WrapErr;
+use mijia::bluetooth::MacAddress;
 use mijia::{MijiaSession, SensorProps};
 use std::collections::HashMap;
 use std::fs::OpenOptions;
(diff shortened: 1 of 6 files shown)


mijia/examples/names.rs
```diff
@@ -0,0 +1,106 @@
...
+fn get_known_sensors() -> Result<Vec<String>, Report> {
+    let sensor_names_contents = std::fs::read_to_string(SENSOR_NAMES_FILE)
+        .wrap_err_with(|| format!("Reading {}", SENSOR_NAMES_FILE))?;
+    let names = toml::from_str::<HashMap<String, String>>(&sensor_names_contents)?;
+
+    if names
+        .keys()
+        .any(|f| f.contains(|c: char| !(c.is_ascii_hexdigit() || c == ':')))
+    {
+        eprintln!("Invalid MAC addresses {:?}", names);
```
alsuren, line 96: hrm. Did we already create a mac address validator that we could use here?

qwandor: `MacAddress::from_str`

mijia/examples/names.rs
```diff
@@ -0,0 +1,106 @@
...
+use mijia::{MijiaSession, SensorProps};
+use std::collections::HashMap;
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::time::Duration;
+use std::{io::stdin, process::exit};
+use tokio::time;
+
+const SCAN_DURATION: Duration = Duration::from_secs(5);
+const SENSOR_NAMES_FILE: &str = "sensor-names.toml";
```
qwandor, line 14: Maybe take this as a command-line flag.

mijia/examples/names.rs
```diff
@@ -0,0 +1,106 @@
...
+            .bt_session
+            .disconnect(&sensor.id)
+            .await
+            .or_else(|_| Ok::<(), Report>(println!("disconnecting failed")))?;
+    }
+
+    Ok(())
+}
+
+fn get_known_sensors() -> Result<Vec<String>, Report> {
```
qwandor, line 87: If you move this to a binary in the `mijia-homie` crate then you can reuse the function there to load this file.

mijia/examples/names.rs
```diff
@@ -0,0 +1,106 @@
...
+    println!("ignoring sensors: {:?}", excludes);
+
+    let (_, session) = MijiaSession::new().await?;
+
+    // Start scanning for Bluetooth devices, and wait a while for some to be discovered.
+    session.bt_session.start_discovery().await?;
+    time::sleep(SCAN_DURATION).await;
+
+    // Get the list of sensors which are currently known, connect those which match the filter and
+    // subscribe to readings.
```
qwandor, line 30: This comment does not appear to be accurate.

mijia/examples/names.rs
qwandor: This should also return a `Vec<MacAddress>` rather than a `Vec<String>`.

mijia/examples/names.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        {
+            let sensor = sensor.clone();
+
+            tokio::task::spawn_blocking(move || {
+                println!("What name do you want to use for {}?", sensor.mac_address);
+                let mut name = String::new();
+                stdin().read_line(&mut name)?;
+                name.truncate(name.trim_end().len());
```
qwandor, line 55: `let name = name.trim_end();`.

mijia/examples/names.rs
```diff
@@ -0,0 +1,106 @@
...
+                Ok::<_, Report>(())
+            })
+            .await??;
+        }
+
+        session
+            .bt_session
+            .disconnect(&sensor.id)
+            .await
+            .or_else(|_| Ok::<(), Report>(println!("disconnecting failed")))?;
```
qwandor, line 81: Log the error.

I think it would be clearer to do `if let Err(e) = ... { println!(...) }` rather than the `or_else` and bogus `?`.

mijia-homie/src/mijia-names.rs
```diff
@@ -0,0 +1,115 @@
...
+    println!(
+        "Finished inspecting all known sensors.\n\n\
+        Some sensors may have been discovered while we were trying to inspect \
+        this batch. Re-run the program to also inspect these new sensors."
+    );
+
+    Ok(())
+}
+
+fn get_known_sensors(sensor_names_filename: &str) -> Result<Vec<MacAddress>, Report> {
```
qwandor, line 104: Can't you just use the existing `config::read_sensor_names` function?

mijia-homie/src/mijia-names.rs
```diff
@@ -0,0 +1,115 @@
...
+    println!("ignoring sensors: {:?}", excludes);
+
+    let (_, session) = MijiaSession::new().await?;
+
+    // Start scanning for Bluetooth devices, and wait a while for some to be discovered.
+    session.bt_session.start_discovery().await?;
+    time::sleep(SCAN_DURATION).await;
+
+    // Get the list of sensors which are currently known, connect those which
+    // are not already known about.
```
qwandor, line 37: Nit: this comment is confusing, because you are using 'known' to mean two different things.

mijia-homie/src/mijia-names.rs
```diff
@@ -0,0 +1,115 @@
...
+                let mut file = OpenOptions::new()
+                    .create(true)
+                    .append(true)
+                    .open(&sensor_names_filename)?;
+                writeln!(
+                    file,
+                    r#""{mac}" = "{name}""#,
+                    mac = sensor.mac_address,
+                    name = name
+                )?;
```
qwandor, line 79: It might be worth adding a helper function for the operation of opening the file and appending one line of the appropriate format to it, as you do that in two different places.

mijia-homie/src/mijia-names.rs
```diff
@@ -0,0 +1,115 @@
...
+        if let Err(e) = session.bt_session.disconnect(&sensor.id).await {
+            log::error!("disconnecting failed: {:?}", e);
+        }
+    }
+
+    println!(
+        "Finished inspecting all known sensors.\n\n\
+        Some sensors may have been discovered while we were trying to inspect \
+        this batch. Re-run the program to also inspect these new sensors."
+    );
```
qwandor, line 99: Optional: you could print a count of how many new sensors were discovered, or even continue asking for names until there are no more.

qwandor (approved):

PR #142, opened 2021-01-28
Add example to fix sensor clocks

qwandor:
Also changes `BluetoothSession::connect` to wait until services have been discovered before returning.

### DIFF
--- a/mijia/examples/fix-clock.rs
+++ b/mijia/examples/fix-clock.rs
@@ -89,16 +89,17 @@ async fn main() -> Result<(), Report> {
             println!("Failed to connect to {}: {:?}", sensor.mac_address, e);
             continue;
         }
-        let last_record = session.get_last_history_record(&sensor.id).await?;
+        let last_record_time = session.get_last_history_record(&sensor.id).await?.time;
         let sensor_time = session.get_time(&sensor.id).await?;
         let now = SystemTime::now();
-        let last_record_utc: DateTime<Utc> = last_record.time.into();
+        let last_record_utc: DateTime<Utc> = last_record_time.into();
         let sensor_time_utc: DateTime<Utc> = sensor_time.into();
         let now_utc: DateTime<Utc> = now.into();
         let offset: SignedDuration = now.duration_since(sensor_time).into();
+        let last_record_offset: SignedDuration = now.duration_since(last_record_time).into();
         println!(
-            "Time: {} (should be {}, offset {:?}), Last stored value: {}",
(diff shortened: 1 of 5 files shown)


mijia/examples/fix-clock.rs
```diff
@@ -0,0 +1,142 @@
...
+use tokio_compat_02::FutureExt;
+
+const SCAN_DURATION: Duration = Duration::from_secs(5);
+/// Only correct clocks which are wrong by more than this amount.
+const MINIMUM_OFFSET: Duration = Duration::from_secs(10);
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// A duration which may be negative.
+#[derive(Clone, Eq, PartialEq)]
+struct SignedDuration {
```
alsuren, line 23: feels like you want something like https://docs.rs/utctimestamp/0.1.4/utctimestamp/struct.TimeDelta.html

mijia/examples/fix-clock.rs
```diff
@@ -0,0 +1,142 @@
...
+        }
+        let last_record = session.get_last_history_record(&sensor.id).await?;
+        let sensor_time = session.get_time(&sensor.id).await?;
+        let now = SystemTime::now();
+        let last_record_utc: DateTime<Utc> = last_record.time.into();
+        let sensor_time_utc: DateTime<Utc> = sensor_time.into();
+        let now_utc: DateTime<Utc> = now.into();
+        let offset: SignedDuration = now.duration_since(sensor_time).into();
+        println!(
+            "Time: {} (should be {}, offset {:?}), Last stored value: {}",
```
alsuren, line 100: (optional) would be interesting to say "last stored value: {}m {}s ago" (which you can calculate using the remote clock)

mijia/examples/fix-clock.rs
alsuren: turns out we're already depending on chrono, so this wouldn't explode our dependency tree too much.

alsuren (approved):
Deeply skeptical about the SignedDuration thing. I guess it's okay for an example though.

mijia/examples/fix-clock.rs
qwandor: Done.

mijia/examples/fix-clock.rs
qwandor: It looks about right in intent, but doesn't have the conversion from `Result<Duration, SystemTimeError>` that is the main point.

PR #149, opened 2021-02-14
Add methods to get information about Bluetooth adapters

qwandor:
This adds `AdapterInfo` with the name, alias, MAC address and status.

### DIFF
--- a/bluez-async/src/adapter.rs
+++ b/bluez-async/src/adapter.rs
@@ -57,10 +57,10 @@ impl AdapterInfo {
     ) -> Result<AdapterInfo, BluetoothError> {
         let mac_address = adapter_properties
             .address()
-            .ok_or_else(|| BluetoothError::RequiredPropertyMissing("Address".to_string()))?;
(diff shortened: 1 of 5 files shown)


bluez-async/src/adapter.rs
```diff
@@ -27,3 +30,101 @@ impl Display for AdapterId {
...
+}
+
+impl AdapterInfo {
+    pub(crate) fn from_properties(
+        id: AdapterId,
+        adapter_properties: OrgBluezAdapter1Properties,
+    ) -> Result<AdapterInfo, BluetoothError> {
+        let mac_address = adapter_properties
+            .address()
+            .ok_or_else(|| BluetoothError::RequiredPropertyMissing("Address".to_string()))?;
```
alsuren, line 60: could this prop take a &'static str or have a constructor that takes &str?

alsuren (approved):
would be nice to use the discovering property to avoid the error when we start_discovery() every few seconds

bluez-async/src/adapter.rs
qwandor: The error type you mean? Done.

PR #151, opened 2021-02-18
Update to latest version of influx_db_client.

qwandor:
This uses Tokio 1.0, so we no longer need tokio-compat-02. It also lets us use rustls rather than openssl.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -820,7 +820,27 @@ dependencies = [
  "indexmap",
  "slab",
  "tokio 0.2.24",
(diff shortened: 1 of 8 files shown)


alsuren (approved):
🎉 all we need now is a rust-based dbus library and we'll be able to cross-compile without docker ;-) 

It's slightly vexing that the clippy lints don't show up on the PRs that introduce them. I wonder if we should pin ourselves to a particular version of clippy, and run the check ourselves as a shell pipeline step.

PR #154, opened 2021-02-27
Add utility to dump historical data to InfluxDB.

qwandor:
I've made this part of the `mijia-homie` crate so that it can share the sensor name reading code.

### DIFF
--- a/mijia-homie/src/mijia-history-influx.rs
+++ b/mijia-homie/src/mijia-history-influx.rs
@@ -69,9 +69,10 @@ async fn write_history(
     name: &str,
     history: Vec<Option<HistoryRecord>>,
 ) -> Result<(), Report> {
-    let points = history.into_iter().filter_map(|record| {
-        record.map(|record| point_for_record(measurement, mac_address, name, &record))
-    });
+    let points = history
+        .into_iter()
+        .flatten()
+        .map(|record| point_for_record(measurement, mac_address, name, &record));
     influxdb_client
         .write_points(points, INFLUXDB_PRECISION, None)
         .await?;
--- a/mijia-homie/src/mijia-history-influx.rs
+++ b/mijia-homie/src/mijia-history-influx.rs
@@ -14,7 +14,6 @@ use tokio::time;
 
 const SCAN_DURATION: Duration = Duration::from_secs(5);
 const INFLUXDB_PRECISION: Option<Precision> = Some(Precision::Milliseconds);
-const MAX_CLOCK_OFFSET: Duration = Duration::from_secs(20 * 60);
 
 #[tokio::main]
 async fn main() -> Result<(), Report> {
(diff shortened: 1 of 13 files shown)


mijia-homie/src/mijia-history-influx.rs
```diff
@@ -0,0 +1,102 @@
+//! Utility program to dump historical data from sensors to InfluxDB.
+
+#[allow(dead_code)]
+mod config;
+mod mijiahistoryconfig;
```
alsuren, line 5: mijia_history_config

mijia-homie/src/mijia-history-influx.rs
```diff
@@ -0,0 +1,102 @@
...
+    for sensor in sensors.iter() {
+        if let Some(name) = names.get(&sensor.mac_address) {
+            log::info!("Connecting to {} ({})", name, sensor.mac_address);
+            if let Err(e) = session.bt_session.connect(&sensor.id).await {
+                log::error!("Failed to connect to {}: {:?}", name, e);
+                continue;
+            }
+
+            log::info!("Reading history...");
+            let history = session.get_all_history(&sensor.id).await?;
```
alsuren, line 46: Assert that the clock is vaguely synchronised || handle clock drift?

mijia-homie/src/mijia-history-influx.rs
```diff
@@ -0,0 +1,102 @@
...
+fn point_for_record(
+    measurement: &str,
+    mac_address: &MacAddress,
+    name: &str,
+    record: &HistoryRecord,
+) -> Point {
+    Point::new(measurement)
+        .add_timestamp(
+            record
+                .time
```
alsuren, line 91: Do these time readings happen to align nicely on the hour, according to the sensor's clock?

alsuren (approved):
Nice work.

This would allow you to backfill a day's data for as many sensors as you want, at a resolution of one reading per hour. I guess this is assuming that the sensor clock is correct to within some reasonable margin (about 10 minutes?). @IkeLutra will be pleased.

If you run it as a cronjob every hour, will influx delicate the readings correctly? I have a feeling that it will, as long as the reading happens on the same millisecond. Wants checking though.

mijia-homie/src/mijia-history-influx.rs
qwandor: Yes.

mijia-homie/src/mijia-history-influx.rs
qwandor: Done.

mijia-homie/src/mijia-history-influx.rs
qwandor: Added a check with a configurable maximum.

qwandor:
> If you run it as a cronjob every hour, will influx delicate the readings correctly? I have a feeling that it will, as long as the reading happens on the same millisecond. Wants checking though.

It appears that InfluxDB does indeed deduplicate the data.

PR #157, opened 2021-03-02
Present boolean as integer for Grafana

kw217:
Grafana can't do anything useful with boolean fields, so this PR adds an additional field `value_int` to any boolean measurements, setting it to `1` if the property is `true` and `0` if the property is `false`. The existing field `value` remains unchanged.

I've also added some documentation about the published measurements to the README.

### DIFF
--- a/homie-influx/README.md
+++ b/homie-influx/README.md
@@ -54,6 +54,17 @@ You may find it helpful to watch the logs to see whether it is managing to conne
 $ sudo journalctl -u homie-influx.service --output=cat --follow
 ```
 
+
+## Format
+
+This service publishes up to six measurements: `integer`, `float`, `boolean`, `string`, `enum`, `color`,
+corresponding to the Homie datatypes. Each message is published as an InfluxDB point with the appropriate
+name and timestamp, the value as a `value` field, and device / node / property info included as tags.
+
+In order to support Grafana clients, boolean points also have an additional `value_int` field,
+which is an integer, 1 for true or 0 for false.
+
+
 ## License
 
 Licensed under either of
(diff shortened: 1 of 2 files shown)


homie-influx/README.md
```diff
@@ -54,6 +54,17 @@ You may find it helpful to watch the logs to see whether it is managing to conne
 $ sudo journalctl -u homie-influx.service --output=cat --follow
 ```
 
+
```
qwandor, line 57: Nit: here and below, one blank line should be enough. (We should probably run prettier in CI.)

qwandor (commented):
Thanks for the PR! I don't actually use boolean fields myself so haven't run into this, but I'm a bit surprised Grafana can't handle booleans. Is there a link to a known Grafana issue we can add for context?

kw217:
Thanks - see e.g. https://github.com/grafana/grafana/issues/8152 and https://github.com/grafana/grafana/issues/24929 (though it looks like there may be a workaround in a recent version - see link in the last issue).

homie-influx/src/influx.rs
```diff
@@ -79,6 +79,12 @@ fn point_for_property_value(
     if let Some(node_type) = node.node_type.to_owned() {
         point = point.add_tag("node_type", Value::String(node_type))
     }
+    if let Some(Datatype::Boolean) = property.datatype {
```
qwandor, line 82: Please add a comment here explaining why this is necessary, with a link to the Grafana issue.

kw217: Done.

homie-influx/README.md
qwandor: Done myself.

kw217:
Thanks!

PR #163, opened 2021-03-11
Store MacAddress as bytes rather than String

qwandor:
Also includes some improvements to error messages and display implementations.

### DIFF
--- a/bluez-async/src/macaddress.rs
+++ b/bluez-async/src/macaddress.rs
@@ -64,3 +64,60 @@ impl FromStr for MacAddress {
         ))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn from_str() {
+        assert_eq!(
+            "11:22:33:44:55:66".parse(),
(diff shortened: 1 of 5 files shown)


bluez-async/src/macaddress.rs
```diff
@@ -0,0 +1,123 @@
...
+use thiserror::Error;
+
+/// An error parsing a MAC address from a string.
+#[derive(Clone, Debug, Error, Eq, PartialEq)]
+#[error("Invalid MAC address '{0}'")]
+pub struct ParseMacAddressError(String);
+
+/// MAC address of a Bluetooth device.
+#[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct MacAddress([u8; 6]);
```
alsuren, line 13: need at least an accessor for this if you want to impl from MacAddress anywhere efficiently

alsuren (approved):
I think it would be better to move this crate into a btleplug workspace repo and share the mac address type as a utils crate, but this pr is still a net win.

bluez-async/src/macaddress.rs
qwandor: Done.

PR #166, opened 2021-03-24
Add support for building to older arm architectures

IkeLutra:
This adds support to build this for older ARM architectures specifically tested it for building for the Raspberry Pi Zero W.

It is essentially a copy of the ARMv7 dockerfile but replacing all references to `armhf` with `armel` which is the debian architecture which supports anything before ARMv7.

_Notes_
- Is ARMv6 the right name? Maybe just ARM?
- This will need building and pushing up the GCR repo if merged.


docker/Dockerfile.debian-buster-armv6
```diff
@@ -0,0 +1,32 @@
...
+RUN dpkg --add-architecture armel && \
+	apt-get update && \
+	apt-get install -y libdbus-1-dev:armel
+
+ENV CARGO_TARGET_ARM_UNKNOWN_LINUX_GNUEABI_LINKER=arm-linux-gnueabi-gcc \
+	CC_arm_unknown_linux_gnueabi=arm-linux-gnueabi-gcc \
+	CXX_arm_unknown_linux_gnueabi=arm-linux-gnueabi-g++ \
+	QEMU_LD_PREFIX=/usr/arm-linux-gnueabi \
+	RUST_TEST_THREADS=1 \
+	PKG_CONFIG_PATH=/usr/lib/arm-linux-gnueabi/pkgconfig
```
qwandor, line 32: Please add a newline at the end of the file.

Cross.toml
```diff
@@ -12,6 +12,12 @@ image = "ghcr.io/qwandor/cross-dbus-debian-buster-armv7:0.2.1-1"
 # context = "./docker"
 # dockerfile = "./docker/Dockerfile.debian-buster-armv7"
 
+[target.arm-unknown-linux-gnueabi]
+# image = "ghcr.io/qwandor/cross-dbus-debian-buster-armv6:TBD"
```
qwandor, line 16: Please uncomment this and give it the same version as the other targets.

qwandor (commented):
Thanks for the PR! It looks good. Please also add the new architecture to `docker/update.sh` and `.github/workflows/package.yml`. I'll push the image.

docker/Dockerfile.debian-buster-armv6
qwandor: Done myself.

Cross.toml
qwandor: Done myself.

PR #167, opened 2021-04-15
Data analysis

anusha-ramdarshan:
exploring the data and making plots


pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[tool.poetry]
```
qwandor, line 1: Can this file and poetry.lock go under notebooks/ rather than in the root directory?

alsuren: I guess we can have a go. Hopefully it won't break nbstripout.

My instinct from my time in python-land is that having .py files in the root of the repo can cause problems later on (when you try to do relative imports), so I always try to have everything nested one level in. I've never used poetry before, but I suspect that moving these two files into notebooks/ will eventually cause us to need an *extra* layer of nesting below that.

Let's try it anyway, and if it works out fine then I will have learned something, and if it gets messy then we can always move things around again.

alsuren (approved):

PR #176, opened 2021-04-28
Fix clippy warnings.

qwandor:


alsuren (approved):
I guess we will need a third pr to update mijia-homie to follow the bluez-async changes?

PR #179, opened 2021-05-08
Export html

alsuren:
The interesting bits are:

Dependencies:
```
ipython-cache = "^0.2.6"
kaleido = "0.2.1"
nbconvert = "^6.0.7"
jupyter-contrib-nbextensions = "^0.5.1"
```

Cache the expensive cell (pickled on disk as notebooks/.cache_magic/dataset/data.txt), so iteration time of re-running the whole notebook is reduced to <4s.
```
import cache_magic
%cache dataset = build_dataset()
```

then set the default renderer to svg

    import plotly.io as pio
    pio.renderers['svg'].engine = 'kaleido'
    pio.renderers.default = 'svg'

If kaleido is not installed then it will send you down a horrible rabbit-hole trying trying to use orca with psutil and requests, which is a big chunk of your life that you will never get back. I managed to force it to attempt kaleido by doing `pio.renderers['svg'].engine = 'kaleido'`, so that it will give users of the notebook correct instructions if they are using it without using this poetry environment.

### DIFF
--- a/notebooks/data_exploration.ipynb
+++ b/notebooks/data_exploration.ipynb
@@ -7,7 +7,10 @@
    "outputs": [],
    "source": [
     "import pandas as pd \n",
-    "import plotly.express as px\n"
+    "import plotly.express as px\n",
+    "import plotly.io as pio\n",
+    "\n",
+    "pio.renderers.default = \"svg\""
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


notebooks/data_exploration.ipynb
```diff
@@ -145,7 +165,7 @@
     "\n",
     "    fig = px.scatter(df, x=\"10min_differential\", y=\"temperature\",color='time_elapsed', \n",
     "                title='temperature vs temperature differential', hover_name='time')\n",
-    "    print(df.shape)\n",
+    "    # print(df.shape)\n",
```
qwandor, line 168: Should this be deleted rather than just commented out?

PR #190, opened 2022-01-19
Handle reconnection better

qwandor:
Rather than relying on a persistent session to keep its subscriptions, the Homie controller should automatically re-subscribe after re-connection. This can be done by making the initial subscription in response to the MQTT `CONACK`.

### DIFF
--- a/homie-controller/src/lib.rs
+++ b/homie-controller/src/lib.rs
@@ -243,12 +243,14 @@ impl HomieController {
         let parts = subtopic.split('/').collect::<Vec<&str>>();
         let event = match parts.as_slice() {
             [device_id, "$homie"] => {
+                // Subscribe to topics even if we already know about the device, because this might
+                // be a reconnection.
+                topics_to_subscribe.push(format!("{}/{}/+", self.base_topic, device_id));
+                topics_to_subscribe.push(format!("{}/{}/$fw/+", self.base_topic, device_id));
+                topics_to_subscribe.push(format!("{}/{}/$stats/+", self.base_topic, device_id));
                 if !devices.contains_key(*device_id) {
                     log::trace!("Homie device '{}' version '{}'", device_id, payload);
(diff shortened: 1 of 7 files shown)


homie-controller/src/lib.rs
```diff
@@ -172,9 +172,8 @@ impl HomieController {
     }
 
     async fn handle_event(&self, incoming: Incoming) -> Result<Option<Event>, PollError> {
-        log::trace!("Incoming: {:?}", incoming);
```
alsuren, line 175: is this trace not useful anymore?

qwandor: It's already logged just up above in `poll`.

qwandor:
TODO: ensure that we do actually subscribe to all the necessary topics for existing devices after a reconnection. It looks like `handle_publish_sync` only subscribes to devices which it thinks are new.

homie-controller/src/lib.rs
```diff
@@ -183,7 +182,13 @@ impl HomieController {
                 }
                 Err(HandleError::Fatal(e)) => return Err(e.into()),
                 Ok(event) => return Ok(event),
+            },
+            Incoming::ConnAck(_) => {
+                // We have connected or reconnected, so make our initial subscription to start
+                // discovering Homie devices.
+                self.start().await?;
```
alsuren, line 189: Should we notify the user of the library that this has happened as well? Feels like it is more important than it was considered to be previously.

alsuren (approved):
fix the bug, and write an integration test to assert that you're properly resubscribed after a reconnect, and then ship it?

homie-controller/src/lib.rs
qwandor: Done.

qwandor:
The integration tests proved more difficult than I hoped so I have left them for now, but fixed the bug and merging this.

PR #192, opened 2022-01-20
Clear devices on reconnect rather than subscribing unconditionally.

qwandor:
Subscribing unconditionally may result in a loop, where getting message for `homie/device/$homie` results in subscribing to `homie/device/+`, which may result in the broker sending the same message again.

### DIFF
--- a/homie-controller/CHANGELOG.md
+++ b/homie-controller/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Bugfixes
+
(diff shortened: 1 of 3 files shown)


qwandor:
I looked at adding a boolean flag to `Device` as we discussed, but it would either need to be public (which is not so nice to expose in the API) or lose the ability for a user of the library to be able to construct a `Device` for testing, which I'd rather not do. So I think just clearing the device map probably is the best thing to do. It shouldn't happen terribly often, and is no worse than your other suggestion of having the entire service restart.

PR #228, opened 2022-09-06
controller: subscribe to $properties before subscribing to their payloads

### LINKED ISSUE #227: controller: properties sometimes have no payload if not updated at runtime
Core issue is this:
https://github.com/alsuren/mijia-homie/blob/e7e0229e7890371838c2b3795e99e7d5a82ee7fb/homie-controller/src/lib.rs#L514-L516

I have a node that updates very rarely. After restart of the controller, my broker publishes the retained payload topic before the node's $properties attribute, causing those payloads to get lost.

I'd suggest not to subscribe to wildcards, instead subscribe explicitly to the attributes, then to the payloads as discovered.

afflux:
Fixes #227

### DIFF
--- a/homie-controller/src/lib.rs
+++ b/homie-controller/src/lib.rs
@@ -794,7 +794,14 @@ mod tests {
 
         // Discover a node on the device.
         publish(&controller, "base_topic/device_id/$nodes", "node_id").await?;
-        expect_subscriptions(&requests_rx, &["base_topic/device_id/node_id/+"]);
+        expect_subscriptions(
+            &requests_rx,
+            &[
+                "base_topic/device_id/node_id/$name",
+                "base_topic/device_id/node_id/$type",
+                "base_topic/device_id/node_id/$properties",
+            ],
+        );
 
         // Discover a property on the node.
         publish(
@@ -805,7 +812,16 @@ mod tests {
         .await?;
         expect_subscriptions(
             &requests_rx,
-            &["base_topic/device_id/node_id/property_id/+"],
+            &[
+                "base_topic/device_id/node_id/property_id",
+                "base_topic/device_id/node_id/property_id/set",
+                "base_topic/device_id/node_id/property_id/$retained",
+                "base_topic/device_id/node_id/property_id/$settable",
+                "base_topic/device_id/node_id/property_id/$format",
+                "base_topic/device_id/node_id/property_id/$unit",
+                "base_topic/device_id/node_id/property_id/$datatype",
+                "base_topic/device_id/node_id/property_id/$name",
+            ],
         );
 
         // No more subscriptions.
--- a/homie-controller/src/lib.rs
+++ b/homie-controller/src/lib.rs
@@ -554,9 +554,6 @@ impl HomieController {
                     && !node_id.starts_with('$')
                     && !property_id.starts_with('$') =>
             {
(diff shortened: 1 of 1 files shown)


afflux:
Some tests probably still need fixing, and adding a regression test case would make sense.

afflux:
Not sure how we'd go for a regression test case, as the subscriptions now changed, so the server wouldn't send the publish packets in the offending order anymore.

homie-controller/src/lib.rs
```diff
@@ -134,6 +134,35 @@ struct PublishResponse {
     topics_to_unsubscribe: Vec<String>,
 }
 
+fn build_node_topics(topics: &mut Vec<String>, base_topic: &str, device_id: &str, node_id: &str) {
```
qwandor, line 137: Nit: as this is adding topics to an existing `Vec`, I'd rather it were called something like `add_node_topics`. (And likewise for `build_property_topics` below.)

qwandor:
This looks like it should work, but I have a couple of concerns:
* It will increase the number of subscriptions the broker has to keep track of by a factor of around 7. I wonder if this will cause any performance issues in large deployments.
* If someone tries to add a feature later that requires other topics, they need to add the new topic in two places (both the code that matches on it, and the code you add here that builds the list). I'm not sure there's any great solution to this, other than some more comments in the source.

One alternative approach I had in mind when I wrote that TODO was to have the code that matches on property values to somehow store values of unknown properties temporarily until the property is discovered, as the issue is really just with the order in which topic values are received. I probably won't have time to look into this in the immediate future though. Do you think it would be feasible? If not, I guess I'll go ahead and merge this as it does fix the bug, but I'd prefer a more performant solution if possible.

(Also, out of curiosity, what are you planning to use this for? If you don't mind saying.)

qwandor:
Oh, I guess you can remove the TODO now too.

afflux:
100% agree with the points regarding naming and comments.
I'm not sure if the number of subscriptions on low-volume topics substantially y affects broker performance, but I'll admit it is something I don't have much experience with. Out of curiosity, I'll try to have a look at the subscription handling of mosquitto.

I agree that maintaining two places for the subtopics is non-ideal, a rust macro could potentially help here, although that might reduce readability.

Regarding the alternative approach, I've thought about that before this PR, and we could store the unknown payloads only for retained messages, so we will only hit that code path on startup. I'd place these in a hashmap with the topic as key, and check for stored values in add_property. (Or we could have a flag that indicates if $properties was received for that node, and only store the unknown payloads if it wasn't yet received). I think it'd require changing the get_mut_property_for. I haven't given a lot of thought if this opens up for a DoS scenario against the broker, where a malicious mqtt client could publish a lot of topics which the controller would have to store, but I guess they could do that with "real" homie properties anyway.

I'm using this homie-controller as the basis for a very lightweight openhab/home-assistant replacement I'm running in my home network out of frustration about their complexity and weight. Haven't really considered to opensource it, though, as it is quite specific to my personal requirements.

I'm out of town right now, will give your approach a try when I get back. Expect a new PR early next week :-)

afflux:
Had a brief look at mosquitto. Seems to be 32byte allocations per subscription (on 64-bit systems), and no significant runtime impact due to hashmap lookups on each topic segment.

PR #231, opened 2022-09-12
controller: handle property values published before  (fixes #227)

### LINKED ISSUE #227: controller: properties sometimes have no payload if not updated at runtime

afflux:
Alternative approach to #228.

Not sure how exactly the `Arc::make_mut` stuff works, I just followed the approach from the `devices` map.
I also don't really like the fact that I have to `format!` the lookup keys when handling `$properties` but I couldn't figure out how else to get tuples as keys working in this case.
Also note that this will indicate Event::PropertyValueChanged to the application even though the property itself wasn't yet discovered, as indicated by the comment around line 545

### DIFF
--- a/homie-controller/src/lib.rs
+++ b/homie-controller/src/lib.rs
@@ -115,6 +115,8 @@ pub struct HomieController {
     /// The set of Homie devices which have been discovered so far, keyed by their IDs.
     // TODO: Consider using Mutex<im::HashMap<...>> instead.
     devices: Mutex<Arc<HashMap<String, Device>>>,
+    /// temporarily holds retained property payloads that were received before their nodes'
+    /// $properties. The stored payloads are consumed when $properties is received.
     early_property_values: Mutex<HashMap<String, String>>,
 }
 
@@ -240,6 +242,7 @@ impl HomieController {
         // our Arc to point to that, so that it is now a unique reference.
         let devices = &mut *self.devices.lock().unwrap();
         let devices = Arc::make_mut(devices);
+
         let early_property_values = &mut *self.early_property_values.lock().unwrap();
 
         // Collect MQTT topics to which we need to subscribe or unsubscribe here, so that the
--- a/homie-controller/src/lib.rs
+++ b/homie-controller/src/lib.rs
@@ -115,7 +115,7 @@ pub struct HomieController {
     /// The set of Homie devices which have been discovered so far, keyed by their IDs.
     // TODO: Consider using Mutex<im::HashMap<...>> instead.
     devices: Mutex<Arc<HashMap<String, Device>>>,
-    early_property_values: Mutex<HashMap<String, String>>,
+    early_property_values: Mutex<Option<HashMap<String, String>>>,
 }
 
 pub struct HomieEventLoop {
@@ -148,7 +148,7 @@ impl HomieController {
             mqtt_client,
             base_topic: base_topic.to_string(),
             devices: Mutex::new(Arc::new(HashMap::new())),
-            early_property_values: Mutex::new(HashMap::new()),
+            early_property_values: Mutex::new(Some(HashMap::new())),
         };
         (controller, HomieEventLoop::new(event_loop))
     }
@@ -426,8 +426,12 @@ impl HomieController {
                 for property_id in properties {
                     if !node.properties.contains_key(property_id) {
                         let mut new_prop = Property::new(property_id);
-                        let key = format!("{}/{}/{}", device_id, node_id, property_id);
-                        new_prop.value = early_property_values.get(&key).cloned();
+
+                        new_prop.value = early_property_values.as_mut().and_then(|map| {
+                            let key = format!("{}/{}/{}", device_id, node_id, property_id);
+                            map.remove(&key)
+                        });
+
                         node.add_property(new_prop);
                         let topic = format!(
                             "{}/{}/{}/{}/+",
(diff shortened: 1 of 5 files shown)


homie-controller/src/lib.rs
```diff
@@ -115,6 +115,7 @@ pub struct HomieController {
     /// The set of Homie devices which have been discovered so far, keyed by their IDs.
     // TODO: Consider using Mutex<im::HashMap<...>> instead.
     devices: Mutex<Arc<HashMap<String, Device>>>,
+    early_property_values: Mutex<Arc<HashMap<String, String>>>,
```
alsuren, line 118: The Mutex<Arc<_>> and .make_mut() dance is to support returning snapshots from the devices() method. I don't think you need to bother for early_property_values. We could probably document this better.

```suggestion
    /// The set of Homie devices which have been discovered so far, keyed by their IDs.
    /// This is Mutex<Arc<_>> rather than Arc<Mutex<_>> so that we can safely hand out
    /// snapshots in the public `devices()` method.
    // TODO: Consider using Mutex<im::HashMap<...>> instead.
    devices: Mutex<Arc<HashMap<String, Device>>>,
    /// Property values from messages that have arrived before the list of node properties is known.
    early_property_values: Mutex<HashMap<String, String>>,
```

homie-controller/src/lib.rs
```diff
@@ -422,7 +426,10 @@ impl HomieController {
                 // Add new properties.
                 for property_id in properties {
                     if !node.properties.contains_key(property_id) {
-                        node.add_property(Property::new(property_id));
+                        let mut new_prop = Property::new(property_id);
+                        let key = format!("{}/{}/{}", device_id, node_id, property_id);
+                        new_prop.value = early_property_values.get(&key).cloned();
```
alsuren, line 431: could we get away with calling .remove() here?

homie-controller/src/lib.rs
```diff
@@ -511,23 +518,43 @@ impl HomieController {
...
+                    Err(_) if publish.retain => {
+                        // temporarily store payloads for unknown properties to prevent
+                        // a race condition when the broker sends out the property
+                        // payloads before $properties
+
+                        early_property_values.insert(subtopic.to_owned(), payload.to_owned());
+
+                        // XXX is this return value correct? should we wait for $properties?
+                        // if so, how to return many PropertyValueChanged from there?
+                        Some(Event::PropertyValueChanged {
```
alsuren, line 547: Good question. I might leave that one for Andrew ;-)

alsuren (commented):
I realise that this is still in draft, but it came up in my emails so I gave it a once-over anyway. Hopefully it answers some of your questions.

homie-controller/src/lib.rs
afflux: Absolutely. Thanks!

homie-controller/src/lib.rs
afflux: Ah, I see. Thanks for clearing that up.

homie-controller/src/lib.rs
```diff
@@ -115,6 +115,7 @@ pub struct HomieController {
     /// The set of Homie devices which have been discovered so far, keyed by their IDs.
     // TODO: Consider using Mutex<im::HashMap<...>> instead.
     devices: Mutex<Arc<HashMap<String, Device>>>,
+    early_property_values: Mutex<Option<HashMap<String, String>>>,
```
qwandor, line 118: It would be useful to have a Rustdoc comment here explaining how this field is used.

homie-controller/src/lib.rs
```diff
@@ -238,6 +240,7 @@ impl HomieController {
         // our Arc to point to that, so that it is now a unique reference.
         let devices = &mut *self.devices.lock().unwrap();
         let devices = Arc::make_mut(devices);
+        let early_property_values = &mut *self.early_property_values.lock().unwrap();
```
qwandor, line 243: Nit: add a blank line before this line, to be clear that the comment above is only referring to the `devices`.

homie-controller/src/lib.rs
qwandor: You could argue either way, but I'd be inclined to wait until the `$properties` come through. The match expression could be changed to return `Vec<Event>` rather than `Option<Event>`.

qwandor:
This looks reasonable to me, just some minor comments. Does it work well for your usecase?

qwandor (commented):
This looks reasonable to me, just some minor comments. Does it work well for your usecase?

homie-controller/src/lib.rs
afflux: This would require changing public API for `poll()`.

qwandor: Fine by me.

afflux:
Note that with 7946a6e826520e5993d4530588120d90d52e8474 I'm not throwing the HashMap away anymore after receiving $properties. See discussion in https://github.com/alsuren/mijia-homie/pull/228#issuecomment-1242675654 on why I think it's not really relevant.

homie-controller/src/lib.rs
afflux: Done. I have to say I still somewhat prefer #228 over this approach, though.

qwandor (approved):

qwandor:
Thanks! This looks good to me now. I'll push a release shortly.

PR #257, opened 2022-12-11
Add service to display sensor values on Raspberry Pi Rainbow Hat

qwandor:
This is a new binary crate that uses `homie-controller` to find all temperature and humidity sensors and display them on a hat attached to a Raspberry Pi. Up to 7 sensor values are displayed graphically on the RGB LEDs, and a single temperature or humidity value on the alphanumeric display. The touch 'buttons' can be used to switch between sensors, switch between temperature and humidty, and adjust the brightness (or turn the display off entirely).

### DIFF
--- a/applehat/Cargo.toml
+++ b/applehat/Cargo.toml
@@ -13,7 +13,6 @@ categories = ["network-programming"]
 color-backtrace = "0.5.1"
 eyre = "0.6.8"
 futures = "0.3.25"
-futures-channel = "0.3.25"
 homie-controller = { version = "0.7.0", path = "../homie-controller" }
 homie-device = { version = "0.7.0", path = "../homie-device" }
 log = "0.4.17"
@@ -23,8 +22,7 @@ rumqttc = "0.18.0"
 rustls = "0.20.7"
 rustls-native-certs = "0.6.2"
 serde_derive = "1.0.137"
-serde = "1.0.148"
+serde = { version = "1.0.148", features = ["derive"] }
 stable-eyre = "0.2.2"
 tokio = { version = "1.22.0", features = ["macros", "rt-multi-thread"] }
 toml = "0.5.9"
-url = { version = "2.3.0", features = ["serde"] }
--- a/applehat/Cargo.toml
+++ b/applehat/Cargo.toml
@@ -26,3 +26,14 @@ serde = { version = "1.0.148", features = ["derive"] }
 stable-eyre = "0.2.2"
 tokio = { version = "1.22.0", features = ["macros", "rt-multi-thread"] }
 toml = "0.5.9"
+
+[package.metadata.deb]
+# $auto doesn't work because we don't build packages in the same container as we build the binaries.
+depends = "libc6"
+section = "net"
+conf-files = ["/etc/applehat/applehat.toml"]
+assets = [
+	["target/release/applehat", "usr/bin/", "755"],
+	["applehat.example.toml", "etc/applehat/applehat.toml", "640"],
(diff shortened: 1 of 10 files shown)


applehat/src/ui.rs
```diff
@@ -0,0 +1,285 @@
...
+            }
+            _ => {}
+        }
+        self.update_display();
+    }
+
+    fn update_button_state(&mut self, new_state: [bool; 3]) {
+        for i in 0..3 {
+            if new_state[i] && !self.button_state[i] {
+                self.button_pressed(i);
```
alsuren, line 163: Are the buttons debounced in hardware or something?

applehat/src/ui.rs
```diff
@@ -0,0 +1,285 @@
...
+        loop {
+            let new_state = [
+                buttons.a.is_pressed(),
+                buttons.b.is_pressed(),
+                buttons.c.is_pressed(),
+            ];
+            ui_state.lock().unwrap().update_button_state(new_state);
+
+            sleep(BUTTON_POLL_PERIOD).await;
+        }
```
alsuren, line 184: :'( killing the planet 100ms at a time. I guess there is no way to wait for a button click on an event driven way?

Does this put you to of the leaderboard on powertop or is there something even more sad running there already?

applehat/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "applehat"
```
alsuren, line 2: Why Applegate? Just because the LEDs are green by default?

alsuren (approved):
Skimmed it and it looks okay. I guess you're already committed to round-tripping to an mqtt server in NZ so all comments about power efficiency can probably be ignored.

applehat/src/ui.rs
qwandor: I guess so? It seems to work, anyway.

applehat/Cargo.toml
qwandor: Applehat not Applegate. Because Heather likes apples, and the Raspberry Pi it's running on is called applepi.

applehat/src/ui.rs
qwandor: Not with the library I'm using. It looks like the underlying hardware does support interrupts though, so maybe I can see about adding support for that.

I haven't checked powertop yet, I'll have a look tonight.

qwandor:
The MQTT broker is in New Jersey not NZ! But yeah, things are going via the Internet.

applehat/src/ui.rs
qwandor: It comes in between 3rd and 8th place, e.g.:

```
Power est.              Usage       Events/s    Category       Description
  3.94 W      3.5 ms/s     995.8        Interrupt      [51] dwc_otg_sim-fiq
  141 mW    250.0 µs/s      35.7        Interrupt      [74] dwc_otg
 28.7 mW      0.9 ms/s       6.9        Process        [PID 640] /usr/bin/telegraf -config /etc/telegraf/telegraf.conf -config-directory /etc/telegraf/telegraf.d
 27.5 mW    148.4 µs/s       6.9        Interrupt      [6] IPI
 25.2 mW      0.8 ms/s       6.1        Timer          hrtimer_wakeup
 25.1 mW    578.4 µs/s       6.1        Timer          tick_sched_timer
 24.0 mW      1.2 ms/s       5.6        Interrupt      [6] tasklet(softirq)
 17.2 mW      2.9 ms/s       3.2        Process        [PID 886] /home/pi/src/mijia-homie/target/debug/applehat
```

PR #360, opened 2023-07-17
Add support for advertising-only sensors

qwandor:


btsensor/src/bthome.rs
```diff
@@ -0,0 +1,637 @@
...
+            Self::Acceleration => "m/s²",
+            Self::Gyroscope => "°/s",
+        }
+    }
+
+    /// The number of spaces to the left to move the decimal point.
+    ///
+    /// In other words, the value stored should be divided by 10 to the power of this number to get
+    /// the actual value.
+    fn decimal_point(self) -> i32 {
```
alsuren, line 437: I would probably call this `denominator` or something, and make it return an integer or float (1.0 or 100.0 or 1000.0)

Cargo.toml
```diff
@@ -2,6 +2,7 @@
 
 members = [
     "applehat",
+    "btsensor",
```
alsuren, line 5: might be better to just name-grab bthome and move the atc parsing somewhere else?

alsuren (commented):
I am apparently watching this repo so closely that I even get notified about draft PRs. I gave it a skim and it looks good so far.

btsensor/src/bthome/v1.rs
```diff
@@ -0,0 +1,657 @@
...
+                | Self::Sound
+                | Self::Tamper
+                | Self::VibrationDetected
+                | Self::WindowOpen
+        )
+    }
+}
+
+/// Attempts to decode the given service data as a BTHome v1 advertisement.
+pub fn decode(mut data: &[u8]) -> Result<Vec<Element>, DecodeError> {
```
alsuren, line 489: (optional suggestion) move this onto Element and call it decode_all()?

btsensor/src/bthome/v2.rs
```diff
@@ -0,0 +1,493 @@
...
+            }
+        }
+    };
+}
+
+generate_element![
+    { 0x00, PacketId, u8, read_u8, "packet ID", "" },
+    { 0x51, Acceleration, u16, read_u16, "acceleration", "m/s²" },
+    { 0x01, Battery, u8, read_u8, "battery", "%"},
+    { 0x12, Co2, u16, read_u16, "CO2", "ppm"},
```
alsuren, line 106: what order are these in? Is that how they're listed in the spec?

alsuren (approved):
🔋📏💡⚖️⚡️🏎️🌡️💦🪫🥶🚪🎤🫨🪟🎛️

btsensor/src/bthome/v2.rs
qwandor: Yes, this is the order from the spec.

btsensor/src/bthome/v1.rs
qwandor: Done.

btsensor/src/bthome.rs
qwandor: Done.

PR #384, opened 2023-08-21
Use btsensor in mijia-homie.

qwandor:
For now the Homie node properties are still hardcoded for BTHome sensors, it assumes they have temperature, humidity and battery percentage, and nothing else. It would be nice to discover properties automatically, but there is no immediate need for that. Only sensors with names configured in `sensor-names.toml` will be included.

### DIFF
--- a/mijia-homie/src/main.rs
+++ b/mijia-homie/src/main.rs
@@ -12,7 +12,7 @@ use futures::stream::StreamExt;
 use futures::TryFutureExt;
 use homie_device::{HomieDevice, Node, Property};
 use itertools::Itertools;
-use log::info;
+use log::{debug, info};
 use mijia::bluetooth::{
     BluetoothError, BluetoothEvent, BluetoothSession, DeviceEvent, DeviceId, MacAddress,
 };
@@ -270,7 +270,12 @@ impl Sensor {
                                             .await?;
                                     }
                                 }
-                                _ => {}
+                                _ => {
+                                    debug!(
+                                        "Skipping unexpected BTHome v1 element {} on {} ({})",
+                                        element, self.mac_address, self.name
+                                    );
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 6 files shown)


mijia-homie/src/main.rs
```diff
@@ -474,6 +657,50 @@ async fn service_bluetooth_event_queue(
 }
 
 async fn handle_bluetooth_event(
```
alsuren, line 659:
```suggestion
async fn handle_broadcast_event(
```

alsuren:
```suggestion
async fn handle_advertisement_event(
```

mijia-homie/src/main.rs
```diff
@@ -283,6 +426,40 @@ struct SensorState {
     min_update_period: Duration,
 }
 
+impl SensorState {
```
alsuren, line 429: I regret not calling this struct SensorStates (initially I read it as the state of a single sensor and wondered why it would have a `add_sensor_if_named()` method.

mijia-homie/src/main.rs
```diff
@@ -196,6 +207,121 @@ impl Sensor {
...
+                                        homie
+                                            .publish_value(
+                                                &node_id,
+                                                Self::PROPERTY_ID_BATTERY,
+                                                value,
+                                            )
+                                            .await?;
+                                    }
+                                }
+                                _ => {}
```
alsuren, line 273:
```suggestion
                                _ => log::debug!("skipping update for {element:?} on {self:?}")
```
or something?

mijia-homie/src/main.rs
qwandor: I called it `handle_bluetooth_event` because it handles a `BluetoothEvent`, i.e. directly from `bluez-async`, as opposed to a `MijiaEvent`. It could handle things other than advertisements in future.

alsuren (approved):

mijia-homie/src/main.rs
qwandor: Good idea, done.

mijia-homie/src/main.rs
qwandor: I've refactored to handle all bluetooth events here, including converting to `MijiaEvent` if appropriate. See what you think.

